PR #4, opened 2024-01-17
[ACI-1529] Optional params parsing

viktorbenei:


internal/config/common/cache_config.go
```diff
@@ -0,0 +1,77 @@
...
+		return createCacheConfig(provider, envProvider("CIRCLE_REPOSITORY_URL"),
+			"", "", "", "")
+	case CIProviderGitHubActions:
+		repoURL := envProvider("GITHUB_SERVER_URL") + "/" + envProvider("GITHUB_REPOSITORY")
+
+		return createCacheConfig(provider, repoURL,
+			"", "", "", "")
+	}
+
+	return CacheConfig{}
```
imrekel, line 76: Is ok that in case of an "unknown" CI provider we just return this?

internal/config/bazel/bazelrc.gotemplate
```diff
@@ -11,3 +11,23 @@ build --remote_header=x-flare-builduser={{ .CIProvider }}
...
+build --bes_header='x-step-id={{ .BitriseStepID }}'
+{{- end }}
+{{- if .BitriseWorkflowName }}
+build --remote_header='x-workflow-name={{ .BitriseWorkflowName }}'
+build --bes_header='x-workflow-name={{ .BitriseWorkflowName }}'
+{{- end }}
+{{- if .BitriseBuildID }}
+build --remote_header='x-build-id={{ .BitriseBuildID }}'
+build --bes_header='x-build-id={{ .BitriseBuildID }}'
+{{- end }}
```
imrekel, line 33: Where do you set the `x-ci-provider`?

imrekel (approved):

internal/config/common/cache_config.go
viktorbenei: It should be. The required envs for auth are required, but any other metadata will only be checked and specified if we know from where to read it.

The name `CacheConfig` is a bit misleading as the struct only contains metadata, which is currently treated as optional.

viktorbenei: If you run the CLI on an unknown CI, this is what the config will be - https://github.com/bitrise-io/bitrise-build-cache-cli/blob/fce35b7d80be2c30d6a41dda3ba08b77f31bba91/internal/config/bazel/bazel_config_test.go#L38
(the `Minimum required params provided` use case, which has the auth infos but no metadata).

PR #5, opened 2024-01-26
Gradle: New Gradle Analytics plugin config

viktorbenei:


internal/config/gradle/gradleconfig_test.go
```diff
@@ -70,106 +101,143 @@ func TestGenerateInitGradle(t *testing.T) {
...
+initscript {
     repositories {
+        mavenLocal()
         mavenCentral()
-        maven {
-            url 'https://jitpack.io'
-        }
+        google()
+        maven(url="https://jitpack.io")
+        maven(url="https://s01.oss.sonatype.org/content/repositories/snapshots/")
```
zsolt-marta-bitrise, line 221: Sonatype is only needed for metrics btw

internal/config/gradle/init.gradle.gotemplate
```diff
@@ -1,73 +1,67 @@
+import io.bitrise.gradle.cache.BitriseBuildCache
```
zsolt-marta-bitrise, line 1: To aoid confusion I'd name this file `init.gradle.kts.gotemplate`, so include that this should be kotlin

internal/config/gradle/gradleconfig.go
```diff
@@ -17,16 +19,17 @@ var (
...
-	CacheVersion    string
-	PushEnabled     bool
-	DebugEnabled    bool
-	ValidationLevel string
-	MetricsEnabled  bool
-	MetricsVersion  string
-	MetricsEndpoint string
-	MetricsPort     int
+	AuthToken           string
+	EndpointURL         string
```
zsolt-marta-bitrise, line 23: Please rename this to something that indicates that this is the cache url, and als othat it contains the port,  like `CacheEndpointWithPort`

internal/config/gradle/gradleconfig.go
```diff
@@ -56,16 +59,17 @@ func GenerateInitGradle(endpointURL, authToken string, analyticsEnabled bool) (s
...
-		MetricsEndpoint: analyticsEndpoint,
-		MetricsPort:     analyticsPort,
+		AuthToken:           authToken,
+		EndpointURL:         endpointURL,
+		CacheVersion:        gradleRemoteBuildCachePluginDepVersion,
+		PushEnabled:         true,
+		DebugEnabled:        true,
+		ValidationLevel:     "warning",
+		MetricsEnabled:      analyticsEnabled,
+		MetricsVersion:      analyticsPluginDepVersion,
```
zsolt-marta-bitrise, line 69: I dont see the change in `analyticsPluginDepVersion`, have you updated that?

internal/config/gradle/gradleconfig.go
```diff
@@ -56,16 +59,17 @@ func GenerateInitGradle(endpointURL, authToken string, analyticsEnabled bool) (s
...
-		EndpointURL:     endpointURL,
-		CacheVersion:    gradleRemoteBuildCachePluginDepVersion,
-		PushEnabled:     true,
-		DebugEnabled:    true,
-		ValidationLevel: "warning",
-		MetricsEnabled:  analyticsEnabled,
-		MetricsVersion:  analyticsPluginDepVersion,
-		MetricsEndpoint: analyticsEndpoint,
-		MetricsPort:     analyticsPort,
+		AuthToken:           authToken,
```
zsolt-marta-bitrise, line 62: Hmm, now I realize that Gradle can only handle PAT as far as I know 🤔

internal/config/gradle/gradleconfig_test.go
```diff
@@ -70,106 +101,143 @@ func TestGenerateInitGradle(t *testing.T) {
...
-                endpoint = 'gradle-analytics.services.bitrise.io'
-                port = 443
-            }
+        registerBuildCacheService(BitriseBuildCache::class.java, BitriseBuildCacheServiceFactory::class.java)
+        remote(BitriseBuildCache::class.java) {
+            endpoint = "grpcs://remote-build-cache.services.bitrise.io"
+            authToken = "AuthT0ken"
+            isEnabled = true
+            isPush = true
+            debug = true
```
zsolt-marta-bitrise, line 133: How are these true if we don't set the args in the test case?

internal/config/gradle/init.gradle.gotemplate
viktorbenei: Done ✅

internal/config/gradle/gradleconfig.go
viktorbenei: Technically this one can work without a port explicitely specified (which is the default how we use it, with just `grpcs://remote-build-cache.services.bitrise.io`). But renamed it to `CacheEndpointURLWithPort`.

internal/config/gradle/gradleconfig.go
viktorbenei: As discussed on Slack we want to keep the existing flows backwards compatible, so the BE will be changed to accept the Bitrise JWT based AuthToken.

internal/config/gradle/gradleconfig_test.go
viktorbenei: Currently it's hardcoded to `true` in the inventory setup code for sake of simplicity.

zsolt-marta-bitrise: Just keep in mind that this is quite verbose and contains a lot of potentially confusing info (albeit still _much_ less verbose than Gradle debug logs :D), so we should not enable it by default for customers

zsolt-marta-bitrise (approved):

PR #14, opened 2024-03-22
Added save gradle diagnostics data

lpusok:


internal/cache/save.go
```diff
@@ -0,0 +1,76 @@
...
+	saveStepID = "save-gradle-build-cache-diagnostics"
+
+	// Cache key template
+	// OS + Arch: to guarantee that stack-specific content (absolute paths, binaries) are stored separately
+	// checksum values:
+	// - `**/*.gradle*`: Gradle build files in any submodule, including ones written in Kotlin (*.gradle.kts)
+	// - `**/gradle-wrapper.properties`: contains exact Gradle version
+	// - `**/gradle.properties`: contains Gradle config values
+	// - `**/gradle/libs.versions.toml`: version catalog file, contains dependencies and their versions
+	key = `{{ .OS }}-{{ .Arch }}-gradle-build-cache-diagnostics-{{ checksum "**/*.gradle*" "**/gradle-wrapper.properties" "**/gradle.properties" "**/gradle/libs.versions.toml" }}`
```
zsolt-marta-bitrise, line 23: Avoid using the `*.gradle` files since those can change, especially because of the reasons we need this feature. Let's really only keep the cache key to the workflow name and the cache limitations like the OS/Arch

internal/cache/save.go
```diff
@@ -0,0 +1,76 @@
...
+		pathChecker:  pathChecker,
+		pathProvider: pathProvider,
+		pathModifier: pathModifier,
+		envRepo:      envRepo,
+	}
+}
+
+func (s GradleDiagnosticOutputSaver) Run(isVerboseMode bool) error {
+	s.logger.Println()
+	s.logger.Printf("Cache key: %s", key)
```
zsolt-marta-bitrise, line 58: Can we do this in debug only mode?

e2e/bitrise.yml
```diff
@@ -0,0 +1,94 @@
...
+        inputs:
+        - content: |
+            #!/bin/env bash
+            set -e
+
+            json_response=$(curl --fail -X POST https://auth.services.bitrise.io/auth/realms/bitrise-services/protocol/openid-connect/token -k \
+                --data "client_id=bitrise-steps" \
+                --data "client_secret=$CACHE_API_CLIENT_SECRET" \
+                --data "grant_type=urn:ietf:params:oauth:grant-type:uma-ticket" \
+                --data "claim_token=eyJhcHBfaWQiOlsiY2FjaGUtc3RlcHMtdGVzdHMiXSwgIm9yZ19pZCI6WyJ0ZXN0LW9yZy1pZCJdLCAiYWJjc19hY2Nlc3NfZ3JhbnRlZCI6WyJ0cnVlIl19" \
```
zsolt-marta-bitrise, line 87: What is this token? I hope this is just for demo. Note: This is a public repo. In any case we'd need envs here

zsolt-marta-bitrise: I tried this and it fails without the `CACHE_API_CLIENT_SECRET`, but still unsure if this is a secret or not

lpusok: It is public in other repos: https://github.com/bitrise-steplib/bitrise-step-save-gradle-cache/blob/2ba01423465d3c5a2bddc2e0942cdb0981c54c74/e2e/bitrise.yml#L122
I remember there was a discussion about it, but not the specifics

internal/cache/restore.go
```diff
@@ -0,0 +1,46 @@
...
+
+	"github.com/bitrise-io/go-steputils/v2/cache"
+	"github.com/bitrise-io/go-utils/v2/command"
+	"github.com/bitrise-io/go-utils/v2/env"
+	"github.com/bitrise-io/go-utils/v2/log"
+)
+
+const (
+	numRestoreRetries = 3
+	restoreStepID     = "restore-gradle-build-cache-diagnostics"
```
zsolt-marta-bitrise, line 14: Why is this required? Can we get this from env?

e2e/bitrise.yml
zsolt-marta-bitrise: Ah ok, thanks! Let's leave a note here to avoid the scare the next time someone finds this :D

internal/cache/restore.go
```diff
@@ -0,0 +1,46 @@
...
+		commandFactory: commandFactory,
+		envRepo:        envRepo,
+	}
+}
+
+func (step GradleOutputDiagnosticsRestorer) Run(isVerboseMode bool) error {
+	if err := cache.NewRestorer(step.envRepo, step.logger, step.commandFactory).Restore(cache.RestoreCacheInput{
+		StepId:         restoreStepID,
+		Verbose:        isVerboseMode,
+		Keys:           []string{key},
```
zsolt-marta-bitrise, line 39: Btw, how is the key expanded from a template? Can we get circleCI envs for the workflow?

e2e/bitrise.yml
```diff
@@ -0,0 +1,95 @@
...
+        - module: app
+        - variant: debug
+    - script:
+        title: Save cache
+        inputs:
+        - content: |-
+            set -ex
+            ../bitrise-build-cache-cli -d save-gradle-output-data
+    - script:
+        title: Delete local Gradle caches
```
imrekel, line 36:
```suggestion
        title: Delete local .gradle folder
```

e2e/bitrise.yml
```diff
@@ -0,0 +1,95 @@
...
+        inputs:
+        - content: |-
+            set -ex
+            ../bitrise-build-cache-cli -d restore-gradle-output-data
+    - script:
+        title: Verify restored folder
+        inputs:
+        - content: |-
+            set -ex
+            if [ ! -d .gradle/caches ]; then
```
imrekel, line 52:
```suggestion
            if [ ! "$(ls -A .gradle)" ]; then
```

Gradle stores the local cache in `$USER_HOME/.gradle/caches` by default (and not in the project's `.gradle` folder) so this will always fail. Verifying that the folder is not empty should work.

e2e/bitrise.yml
```diff
@@ -0,0 +1,95 @@
...
+        inputs:
+        - path: ./_tmp
+    # - set-java-version@1:
+    - android-build:
+        title: Build APK
+        inputs:
+        - module: app
+        - variant: debug
+    - script:
+        title: Save cache
```
imrekel, line 30:
```suggestion
        title: Save gradle output
```

e2e/bitrise.yml
```diff
@@ -0,0 +1,95 @@
...
+            set -ex
+            ../bitrise-build-cache-cli -d save-gradle-output-data
+    - script:
+        title: Delete local Gradle caches
+        inputs:
+        - content: |-
+            set -ex
+            sudo rm -rf .gradle
+    - script:
+        title: Restore cache
```
imrekel, line 42:
```suggestion
        title: Restore Gradle output data
```

internal/cache/restore.go
```diff
@@ -32,15 +33,40 @@ func NewGradleDiagnosticOuptutRestorer(
...
-	return nil
+	foundRestoredData, err := folderExistsNotEmpty(".gradle")
+	if err != nil {
+		return false, fmt.Errorf("failed to check if Gradle output data was restored: %w", err)
+	}
+
+	return foundRestoredData, nil
+}
+
+func folderExistsNotEmpty(folderPath string) (bool, error) {
```
zsolt-marta-bitrise, line 54: Do we have anything like this in the utils libs we're pulling in?

zsolt-marta-bitrise (approved):

PR #27, opened 2024-06-11
[ACI-1586] Add new inputs

lpusok:
Added debug, validation level, pus hand endpoint inputs.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -23,3 +23,4 @@ linters:
   - testpackage
   - mnd
   - gomnd
+  - err113 # ensble use of dynamic errors fmt.Errorf("error: %s", str)
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -23,4 +23,3 @@ linters:
   - testpackage
   - mnd
   - gomnd
-  - err113 # ensble use of dynamic errors fmt.Errorf("error: %s", str)
--- a/cmd/enableForGradle.go
+++ b/cmd/enableForGradle.go
@@ -15,7 +15,8 @@ import (
 
 const gradleHomeNonExpanded = "~/.gradle"
 
-var ( //nolint:gochecknoglobals
+//nolint:gochecknoglobals
+var (
 	paramIsGradleMetricsEnabled bool
 	paramIsPushEnabled          bool
 	paramValidationLevel        string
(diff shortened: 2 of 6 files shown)


.golangci.yaml
```diff
@@ -23,3 +23,4 @@ linters:
   - testpackage
   - mnd
   - gomnd
+  - err113 # ensble use of dynamic errors fmt.Errorf("error: %s", str)
```
zsolt-marta-bitrise, line 26: Why do you need `fmt.Errorf` for those cases, and why not use `fmt.Sprintf`? Errorf should be used for `fmt.Errorf("call api: %w", err)`

internal/config/gradle/gradleconfig_test.go
```diff
@@ -77,9 +87,14 @@ func TestGenerateInitGradle(t *testing.T) {
...
-				endpointURL:         "grpcs://remote-build-cache.services.bitrise.io",
-				authToken:           "AuthT0ken",
-				metricsEnabled:      true,
+				endpointURL: "grpcs://remote-build-cache.services.bitrise.io",
+				authToken:   "AuthT0ken",
+				userPrefs: Preferences{
+					IsPushEnabled:        true,
+					CacheLevelValidation: CacheValidationLevelWarning,
+					IsAnalyticsEnabled:   true,
+					IsDebugEnabled:       true,
```
zsolt-marta-bitrise, line 96: Please add a test case for debug = false and push = false

cmd/enableForGradle.go
```diff
@@ -71,9 +80,18 @@ func enableForGradleCmdFn(logger log.Logger, gradleHomePath string, envProvider
 
 	// Optional configs
 	// EndpointURL
-	endpointURL := common.SelectEndpointURL(envProvider("BITRISE_BUILD_CACHE_ENDPOINT"), envProvider)
+	endpointURL := common.SelectEndpointURL(paramRemoteCacheEndpoint, envProvider)
 	logger.Infof("(i) Build Cache Endpoint URL: %s", endpointURL)
+	logger.Infof("(i) paramIsPushEnabled: %t", paramIsPushEnabled)
+	logger.Infof("(i) paramValidationLevel: %s", paramValidationLevel)
 	logger.Infof("(i) paramIsGradleMetricsEnabled: %t", paramIsGradleMetricsEnabled)
+	logger.Infof("(i) isDebugLogMode: %t", isDebugLogMode)
```
zsolt-marta-bitrise, line 88: Where is this populated?

.golangci.yaml
lpusok: This linter also prevents  `errors.New("")` for example.

cmd/enableForGradle.go
lpusok: Here: https://github.com/bitrise-io/bitrise-build-cache-cli/blob/38f5096b3ce65f6637a2eb996001c61b006e23ca/cmd/root.go#L34

.golangci.yaml
zsolt-marta-bitrise: I seriously doubt that 🤔 Are you sure?

zsolt-marta-bitrise (approved):

.golangci.yaml
zsolt-marta-bitrise: I saw you reverted this, so approving

PR #44, opened 2024-08-21
[ACI-2730] Operation analytics

imrekel:
here is a test build: https://app.bitrise.io/build/28ca04dd-3ec4-4b11-9494-6127b8fe82d6

note: the transfer_size column has been migrated to bigint in the service DB

### DIFF
--- a/cmd/analytics.go
+++ b/cmd/analytics.go
@@ -1,11 +1,45 @@
 package cmd
 
 import (
-	xa "github.com/bitrise-io/bitrise-build-cache-cli/internal/analytics"
-	"github.com/bitrise-io/bitrise-build-cache-cli/internal/xcode"
-	"github.com/google/uuid"
+	"encoding/json"
+	"fmt"
 	"time"
+
+	"github.com/google/uuid"
+
+	xa "github.com/bitrise-io/bitrise-build-cache-cli/internal/analytics"
+	"github.com/bitrise-io/bitrise-build-cache-cli/internal/config/common"
+	"github.com/bitrise-io/bitrise-build-cache-cli/internal/consts"
+	"github.com/bitrise-io/bitrise-build-cache-cli/internal/xcode"
+	"github.com/bitrise-io/go-utils/v2/log"
 )
+
+func sendCacheOperationAnalytics(op *xa.CacheOperation, cmdError error, logger log.Logger, authConfig common.CacheAuthConfig) {
+	op.Duration = int(time.Since(op.StartedAt).Milliseconds())
+
+	if cmdError != nil {
+		errStr := cmdError.Error()
+		op.Error = &errStr
+	}
+
+	xaClint, clientErr := xa.NewClient(consts.AnalyticsServiceEndpoint, authConfig.AuthToken, logger)
+	if clientErr != nil {
+		logger.Warnf("Failed to create Xcode Analytics Service client: %s", clientErr)
+	} else {
+		logger.Debugf("Sending cache operation to Xcode Analytics Service: %v", func() string {
+			payload, err := json.Marshal(op)
+			if err != nil {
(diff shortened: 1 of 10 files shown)


cmd/analytics.go
```diff
@@ -0,0 +1,102 @@
...
+	"github.com/google/uuid"
+
+	xa "github.com/bitrise-io/bitrise-build-cache-cli/internal/analytics"
+	"github.com/bitrise-io/bitrise-build-cache-cli/internal/config/common"
+	"github.com/bitrise-io/bitrise-build-cache-cli/internal/consts"
+	"github.com/bitrise-io/bitrise-build-cache-cli/internal/xcode"
+	"github.com/bitrise-io/go-utils/v2/log"
+)
+
+func sendCacheOperationAnalytics(op *xa.CacheOperation, cmdError error, logger log.Logger, authConfig common.CacheAuthConfig) {
```
zsolt-marta-bitrise, line 16: Why pass as pointer? The struct is small and has no performance impact, but now this method has a random side effect

cmd/analytics.go
```diff
@@ -0,0 +1,102 @@
...
+	op.Duration = int(time.Since(op.StartedAt).Milliseconds())
+
+	if cmdError != nil {
+		errStr := cmdError.Error()
+		op.Error = &errStr
+	}
+
+	xaClint, clientErr := xa.NewClient(consts.AnalyticsServiceEndpoint, authConfig.AuthToken, logger)
```
zsolt-marta-bitrise, line 24:
```suggestion
	xaClient, clientErr := xa.NewClient(consts.AnalyticsServiceEndpoint, authConfig.AuthToken, logger)
```

cmd/analytics.go
```diff
@@ -0,0 +1,102 @@
...
+	if clientErr != nil {
+		logger.Warnf("Failed to create Xcode Analytics Service client: %s", clientErr)
```
zsolt-marta-bitrise, line 26: Why not return error?

zsolt-marta-bitrise: Same for below, would simplify the conditions and logging, and improve readability

internal/analytics/types.go
```diff
@@ -0,0 +1,37 @@
...
+	FilesFailed      int `json:"filesFailed"`
+	FilesMissing     int `json:"filesMissing"`
+	TotalFiles       int `json:"totalFiles"`
+}
+
+type CacheOperation struct {
+	OperationID   string    `json:"operationId"`
+	OperationType string    `json:"operationType"`
+	StartedAt     time.Time `json:"startedAt"`
+	Duration      int       `json:"durationMs"`
```
zsolt-marta-bitrise, line 22: Please call this DurationMilliseconds or MS to include that this is milliseconds

internal/config/common/auth_test.go
```diff
@@ -0,0 +1,74 @@
+package common
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestReadAuthConfigFromEnvironments(t *testing.T) {
```
zsolt-marta-bitrise, line 9: Yeah we should soon start adding tests :D Thanks 👍🏻

internal/analytics/types.go
imrekel: ok (btw this is straight from the Analytics Service code, so we should change it there too then)

cmd/analytics.go
zsolt-marta-bitrise: You're setting properties of a struct passed as a pointer

imrekel: indeed 😄

zsolt-marta-bitrise (approved):

PR #50, opened 2024-08-30
[ACI-2767] Handling existing files, logging improvements

imrekel:
- new parameter to force overwrite existing files
- limit the number of download errors logged
- log the largest 10 files

### DIFF
--- a/cmd/restoreXcodeDerivedDataFiles.go
+++ b/cmd/restoreXcodeDerivedDataFiles.go
@@ -249,8 +249,8 @@ func logCacheMetadata(md *xcode.Metadata, logger log.Logger, isDebugLogMode bool
 			sortedDDFiles = sortedDDFiles[:10]
 		}
 
+		logger.Debugf("  Largest files:")
 		for i, file := range sortedDDFiles {
-			logger.Infof("  Largest files:")
 			logger.Debugf("    %d. %s (%s)", i+1, file.Path, humanize.Bytes(uint64(file.Size)))
 		}
 	}
--- a/cmd/restoreXcodeDerivedDataFiles.go
+++ b/cmd/restoreXcodeDerivedDataFiles.go
@@ -6,8 +6,11 @@ import (
 	"fmt"
 	"os"
 	"os/user"
+	"sort"
 	"strings"
 	"time"
+
+	"github.com/dustin/go-humanize"
 
 	xa "github.com/bitrise-io/bitrise-build-cache-cli/internal/analytics"
 	"github.com/bitrise-io/bitrise-build-cache-cli/internal/build_cache/kv"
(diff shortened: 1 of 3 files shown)


cmd/restoreXcodeDerivedDataFiles.go
```diff
@@ -85,10 +91,12 @@ func init() {
 	if err := restoreXcodeDerivedDataFilesCmd.MarkFlagRequired("project-root"); err != nil {
 		panic(err)
 	}
+	restoreXcodeDerivedDataFilesCmd.Flags().Bool("force-overwrite", false, "If set, the command will try to overwrite existing files during restoring the cache even if the permissions do not allow it")
```
zsolt-marta-bitrise, line 94: Can you pls name the flag something that indicates that it overwrites only the files, and not the cache? like `force-overwrite-files`

internal/xcode/download.go
```diff
@@ -40,6 +44,24 @@ func downloadFile(ctx context.Context, client *kv.Client, filePath, key string,
 	if fileMode == 0 {
 		fileMode = 0666
 	}
+
+	if fileInfo, err := os.Stat(filePath); err == nil {
```
zsolt-marta-bitrise, line 48: Please early return if there is an error to reduce indentation

internal/xcode/download_multi.go
```diff
@@ -75,7 +79,9 @@ func DownloadCacheFilesFromBuildCache(ctx context.Context, dd FileGroupInfo, kvC
 
 				filesMissing.Add(1)
 			case err != nil:
-				logger.Errorf("Failed to download file %s with error: %v", file.Path, err)
+				if filesFailedToDownload.Load() < int32(maxLoggedDownloadErrors) {
```
zsolt-marta-bitrise, line 82: Btw, should we stop the download too in this case? It doesn't really make sense to continue trying and it slows down the build significantly

zsolt-marta-bitrise: I meant if we exceed the max errors

internal/xcode/download.go
```diff
@@ -40,6 +44,24 @@ func downloadFile(ctx context.Context, client *kv.Client, filePath, key string,
...
+		ownerWritable := (fileInfo.Mode().Perm() & 0200) != 0
```
zsolt-marta-bitrise, line 49: Why are we checking the owner writable flag? Is the current user always going to be the owner?

internal/xcode/download_multi.go
imrekel: We can think about it - as of now, this could mean that we just cannot overwrite existing files (and no force overwrite is enabled). So in theory this could still work.

internal/xcode/download.go
imrekel: I think it should be, the Bitrise build is running by the same user (`vagrant`) - unless somone does some hacking with switching users and such, I don't see why it would change

internal/xcode/download.go
imrekel: what to early return here? The inner block (your next comment) is ok, but I can't see how to reduce nesting even more (there is no clean way to "break" from nested ifs in go)

zsolt-marta-bitrise: Ah ok this only checks if the file exists, this is just a special case of the method, not a requirement.

Please then extract to a separate method to reduce the complexity (it's getting hard to read)

zsolt-marta-bitrise (approved):
not a blocker but please consider [this](https://github.com/bitrise-io/bitrise-build-cache-cli/pull/50/files#r1738587696)

PR #51, opened 2024-09-04
[ACI-2748] Send cache API metadata + limit logging missing files

imrekel:


cmd/client.go
```diff
@@ -29,12 +29,13 @@ func createKVClient(ctx context.Context, authConfig common.CacheAuthConfig, envP
...
-		ClientName:  "kv",
-		AuthConfig:  authConfig,
-		Logger:      logger,
+		UseInsecure:         insecureGRPC,
+		Host:                buildCacheHost,
+		DialTimeout:         5 * time.Second,
+		ClientName:          "kv",
+		AuthConfig:          authConfig,
+		Logger:              logger,
+		CacheConfigMetadata: common.NewCacheConfigMetadata(envProvider),
```
zsolt-marta-bitrise, line 38: Don't forget we'll need to check/update [this](https://github.com/bitrise-io/bitrise-build-cache-cli/blob/4b51cd60c7cc141d99d6a3a1e19963f2fa014b74/internal/config/common/cache_config.go#L68)

zsolt-marta-bitrise (approved):

cmd/client.go
imrekel: What do you mean? (I checked and the current env vars are valid)

PR #52, opened 2024-09-06
[NOJIRA] Send cache operation ID in cache API calls + fix restore analytics

imrekel:


cmd/restoreXcodeDerivedDataFiles.go
```diff
@@ -97,13 +97,15 @@ func init() {
...
-	kvClient, err := createKVClient(ctx, authConfig, envProvider, logger)
+	op := newCacheOperation(startT, xa.OperationTypeDownload, envProvider)
+	kvClient, err := createKVClient(ctx, op.OperationID, authConfig, envProvider, logger)
 	if err != nil {
 		return nil, fmt.Errorf("create kv client: %w", err)
 	}
+	logger.Infof("(i) Cache operation ID: %s", op.OperationID)
 
 	cacheKeyType, cacheKey, err := downloadMetadata(ctx, cacheMetadataPath, providedCacheKey, kvClient, logger, envProvider)
-	op := newCacheOperation(startT, xa.OperationTypeUpload, cacheKey, envProvider)
```
imrekel, line 106: oops... (`xa.OperationTypeUpload)

cmd/deleteXcodeDerivedData.go
```diff
@@ -70,7 +70,7 @@ func deleteXcodeDerivedDataCmdFn(ctx context.Context, providedCacheKey string, u
 	}
 	logger.Infof("(i) Cache key: %s", cacheKey)
 
-	kvClient, err := createKVClient(ctx, authConfig, envProvider, logger)
```
zsolt-marta-bitrise, line 73: Why don't we create an operation ID here too?

cmd/analytics.go
```diff
@@ -33,12 +33,11 @@ func sendCacheOperationAnalytics(op xa.CacheOperation, logger log.Logger, authCo
 	return nil
 }
 
-func newCacheOperation(startT time.Time, operationType, cacheKey string, envProvider func(string) string) *xa.CacheOperation {
+func newCacheOperation(startT time.Time, operationType string, envProvider func(string) string) *xa.CacheOperation {
```
zsolt-marta-bitrise, line 36: When can we have an operation without a cache key? What's the benefit of setting it in a separate step after this call?

cmd/deleteXcodeDerivedData.go
imrekel: We don't really have a "delete" operation defined yet - the whole "operation" concept is now focuseed on upload/download with their parameters and such. So I wouldn't start assigning IDs to this "arbitrary" delete operation as we don't support it outside of internal testing. I can imagine we will want a proper delete oepration but then that's need to be thought through.

cmd/analytics.go
imrekel: it's because we need an operation ID before we have cache key and currently operation ID is created with the operation (the actual cache key is returned after downloading the metadata due to the fallback mechanism)

```
op := newCacheOperation(startT, xa.OperationTypeDownload, envProvider)
	kvClient, err := createKVClient(ctx, op.OperationID, authConfig, envProvider, logger)
	if err != nil {
		return nil, fmt.Errorf("create kv client: %w", err)
	}
	logger.Infof("(i) Cache operation ID: %s", op.OperationID)

	cacheKeyType, cacheKey, err := downloadMetadata(ctx, cacheMetadataPath, providedCacheKey, kvClient, logger, envProvider)
```

an alternative would be generating the operation ID before having an operation and then perform the metadata download and create the operation after that...

zsolt-marta-bitrise: Ok yeah, in this case we have a cache operation with unknown key, makes sense

zsolt-marta-bitrise (approved):

PR #53, opened 2024-09-10
feat: ACI-2805 Add flag to follow xcode symlinks

zsolt-marta-bitrise:


cmd/saveXcodeDerivedDataFiles.go
```diff
@@ -49,7 +50,18 @@ var saveXcodeDerivedDataFilesCmd = &cobra.Command{
...
-		op, cmdError := saveXcodeDerivedDataFilesCmdFn(cmd.Context(), authConfig, CacheMetadataPath, projectRoot, cacheKey, ddPath, xcodeCachePath, logger, tracker, startT, os.Getenv)
+		op, cmdError := saveXcodeDerivedDataFilesCmdFn(cmd.Context(),
+			authConfig,
+			CacheMetadataPath,
+			projectRoot,
+			cacheKey,
+			ddPath,
+			xcodeCachePath,
+			followSymlinks,
+			logger,
```
zsolt-marta-bitrise, line 61: Reminder https://bitrise.atlassian.net/browse/ACI-2397

internal/xcode/file_group_info.go
```diff
@@ -103,29 +110,60 @@ func collectFileGroupInfo(cacheDirPath string, rootDir string, collectAttributes
...
 
 		return nil
 	}
 
-	if inf.Mode()&os.ModeSymlink != 0 {
-		logger.Debugf("Skipping symbolic link: %s", path)
-
-		return nil
+	if fileInfo.Mode()&os.ModeSymlink != 0 {
+		return collectSymlink(path, fgi, followSymlinks, rootDir, logger)
```
imrekel, line 166: I don't get it, if a file is symlink, we call `collectSymlink` on it. But `collectSymlink` will call `collectFileMetadata` on the file if it's not a directory. Do I miss something on how the circular method calls are broken?

zsolt-marta-bitrise: `collectSymlink()` will call `resolveSymlink()` to resolve the symlink and collect the file info on that target. The resolution is called recursively until we find a regular file, so `collectFileMetadata()` is called again with the target files only

internal/xcode/file_group_info.go
```diff
@@ -197,6 +286,42 @@ func setAttributes(path string, attributes map[string]string) error {
 	return nil
 }
 
+func restoreSymlink(symlink SymlinkInfo, forceOverwrite bool, logger log.Logger) bool {
+	fileInfo, err := os.Lstat(symlink.Path)
+	if err == nil && fileInfo.Mode()&os.ModeSymlink != 0 {
+		if !forceOverwrite {
```
imrekel, line 292: just a note: currently we overwrite regular files by default (even if forceOverwrite=false) - forceOverwrite=true means that we even try to overWrite files with read only file mode

so I'd remove this check and overwrite symlinks by default too - we will need a new config option to skip existing files and if that's set then I'd skip symlinks too

cmd/saveXcodeDerivedDataFiles.go
```diff
@@ -86,10 +98,21 @@ func init() {
 		panic(err)
 	}
 	saveXcodeDerivedDataFilesCmd.Flags().String("xcodecache-path", "", "Path to the Xcode cache directory folder to be saved. If not set, it will not be uploaded.")
+	saveXcodeDerivedDataFilesCmd.Flags().Bool("follow-symlinks", false, "Follow symlinks when calculating metadata (default: false)")
```
imrekel, line 101:
```suggestion
	saveXcodeDerivedDataFilesCmd.Flags().Bool("follow-symlinks", false, "Follow symlinks when calculating metadata and saving files to the cache (default: false)")
```
```suggestion
	saveXcodeDerivedDataFilesCmd.Flags().Bool("follow-symlinks", false, "Follow symlinks when calculating metadata (default: false)")
```

imrekel (approved):

PR #54, opened 2024-09-17
feat: ACI-2805 Add flag to skip SPM

zsolt-marta-bitrise:


internal/xcode/file_group_info.go
```diff
@@ -211,6 +227,17 @@ func collectFileMetadata(path string, fileInfo fs.FileInfo, isDirectory bool, fg
 		return nil
 	}
 
+	if skipSPM {
+		relPath, _ := filepath.Rel(rootPath, path)
+		parts := strings.Split(filepath.ToSlash(relPath), "/")
+
+		// Checking for SourcePackages/* or */SourcePackages/* under the directory
+		if len(parts) >= 1 && parts[0] == "SourcePackages" ||
+			len(parts) >= 2 && parts[1] == "SourcePackages" {
```
imrekel, line 236: So this is always directly under DD or at level 1?

imrekel (approved):

internal/xcode/file_group_info.go
zsolt-marta-bitrise: Can be both, if DD is directly specified, it will be in the root, if it's the generic default DD directory with each projects having their own subfolder it's in 2. level

PR #55, opened 2024-09-25
[ACI-2869] Fix duplicate blob uploads

imrekel:
- delete is a noop if key doesn't exist
- we can delete the blob in advance as the `uploadCacheFileToBuildCache` will run eventually and that will either download the blob or it will be skipped with an error after the retries

### DIFF
--- a/internal/xcode/upload_multi.go
+++ b/internal/xcode/upload_multi.go
@@ -69,6 +69,7 @@ func UploadCacheFilesToBuildCache(ctx context.Context, dd FileGroupInfo, kvClien
 	for _, file := range dd.Files {
 		mutex.Lock()
 		_, ok := missingBlobs[file.Hash]
+		delete(missingBlobs, file.Hash) // Remove the blob from the list of missing blobs as it's being uploaded
 		mutex.Unlock()
 		if !ok {
 			continue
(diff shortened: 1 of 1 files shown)


internal/xcode/upload_multi.go
```diff
@@ -69,6 +69,7 @@ func UploadCacheFilesToBuildCache(ctx context.Context, dd FileGroupInfo, kvClien
 	for _, file := range dd.Files {
 		mutex.Lock()
 		_, ok := missingBlobs[file.Hash]
+		delete(missingBlobs, file.Hash) // Remove the blob from the list of missing blobs as it's being uploaded
```
zsolt-marta-bitrise, line 72: If a file fails to upload then all files having the same blob will be missing and there will be no retries but seems like an edge case and we already handle per-file failures

zsolt-marta-bitrise (approved):

internal/xcode/upload_multi.go
imrekel: I agree - I don't think we need to take this case into account ( = "there are multiple files sharing a blob in an uploads and then one file for the blob fails completely but we expect that another file for the blob might succeed")

PR #56, opened 2024-09-27
[ACI-2800] Option to skip existing files when restoring the cache

imrekel:


internal/xcode/download_multi.go
```diff
@@ -44,10 +44,19 @@ func DownloadCacheFilesFromBuildCache(ctx context.Context, dd FileGroupInfo, kvC
 	var filesMissing atomic.Int32
 	var filesFailedToDownload atomic.Int32
 	var downloadSize atomic.Int64
+	var skippedFiles atomic.Int32
 
 	var wg sync.WaitGroup
 	semaphore := make(chan struct{}, 20) // Limit parallelization
 	for _, file := range dd.Files {
+		if skipExisting {
+			if _, err := os.Stat(file.Path); err == nil {
```
zsolt-marta-bitrise, line 53: We now also have some file stat logic in `downloadFile()`, did you think about merging these so that we have all the logic related to files in a single place?

zsolt-marta-bitrise (approved):

PR #57, opened 2024-09-27
feat: ACI-2926 Generate config cache compatible config

zsolt-marta-bitrise:


internal/config/gradle/initd.gradle.kts.gotemplate
```diff
@@ -53,12 +56,6 @@ rootProject {
             {{- if .CacheConfigMetadata.BitriseAppID }}
             appSlug.set("{{ .CacheConfigMetadata.BitriseAppID }}")
             {{- end }}
-            {{- if .CacheConfigMetadata.BitriseBuildID }}
```
zsolt-marta-bitrise, line 56: The  analytics plugin will read these (in a config cache compatible way)

PR #59, opened 2024-10-04
feat: ACI-2928 Create save/restore config cache commands

zsolt-marta-bitrise:


cmd/saveGradleConfigurationCache.go
```diff
@@ -0,0 +1,149 @@
...
+	if err != nil {
+		return fmt.Errorf("create kv client: %w", err)
+	}
+
+	cache := gradle.NewCache(logger, envProvider, kvClient)
+
+	var cacheKey string
+	if providedCacheKey == "" {
+		logger.Infof("(i) Cache key is not explicitly specified, setting it based on the current Bitrise app's slug and git branch...")
+		if cacheKey, err = cache.GetCacheKey(gradle.CacheKeyParams{}); err != nil {
```
imrekel, line 86: this key will be the same for an xcode deriveddata cache executed on the same app/branch - I wonder if we want to make the keys unique based on the type of data stored under them

zsolt-marta-bitrise: The implementation is duplicated in `gradle` package, this is a call on the struct, so it will be different. See [Here](https://github.com/bitrise-io/bitrise-build-cache-cli/pull/59/files#diff-6377a19e855ca595bcb521e6b0f3ea402c2d61a0675f9ab8ec70291170e8f2b7R24)

imrekel (approved):

PR #68, opened 2024-11-29
Make it easier to update gradle verification reference

lpusok:
- Created a step bundle and WF (`check_gradle_verification`) that can be used to update verification-metadata.xml.
This is now included when running tests, to make sure it is kept up-to-date.
- Added deps that were missing from verification-metadata.xml. Despite using the same gradle version, when running in CI there were minor differences.


bitrise.yml
```diff
@@ -79,9 +54,10 @@ workflows:
     - deploy-to-bitrise-io@2: {}
     meta:
       bitrise.io:
-        machine_type_id: g2-m1.8core
-        stack: osx-xcode-edge
+        stack: osx-xcode-16.1.x
   test:
+    after_run:
```
zsolt-marta-bitrise, line 59: Since `check_gradle_verification` contains only `bundle::generate_gradle_verification_reference: { }`, and that is also explicitly added here, it will be run twice. Why do we need this after_run?

bitrise.yml
```diff
@@ -113,5 +93,57 @@ workflows:
...
             set -ex
-            bitrise run --config e2e/bitrise.yml test_gradle_output_diagnostics 
+            bitrise run --config e2e/bitrise.yml test_gradle_output_diagnostics
+
+  check_gradle_verification:
+    steps:
+    - bundle::generate_gradle_verification_reference: { }
+
+step_bundles:
+  generate_gradle_verification_reference:
```
zsolt-marta-bitrise, line 103: Since this also performs a check I'd call this `check_gradle_verification_reference`

lpusok: renamed.

zsolt-marta-bitrise (approved):

PR #73, opened 2024-12-11
fix: Use client name specific for build tools

zsolt-marta-bitrise:


internal/build_cache/kv/methods.go
```diff
@@ -80,7 +80,7 @@ func (c *Client) Get(ctx context.Context, name string) (io.ReadCloser, error) {
 }
 
 func (c *Client) Delete(ctx context.Context, name string) error {
-	resourceName := fmt.Sprintf("%s/%s", c.clientName, name)
```
zsolt-marta-bitrise, line 83: Changing these would invalidate caches

cmd/client.go
```diff
@@ -36,7 +42,7 @@ func createKVClient(ctx context.Context,
 		UseInsecure:         insecureGRPC,
 		Host:                buildCacheHost,
 		DialTimeout:         5 * time.Second,
-		ClientName:          "kv",
+		ClientName:          clientName,
```
zsolt-marta-bitrise, line 45: Reused the existing field that wasn't doing much

Badlazzor (approved):

PR #81, opened 2025-01-14
feat: ACI-3311 Generate dependency matrix 

zsolt-marta-bitrise:


bitrise.yml
```diff
@@ -152,3 +167,81 @@ step_bundles:
...
+                
+                pandoc --version
+      - script:
+          title: Install go versions
+          inputs:
+            - content: |
+                #!/bin/bash
+                set -ex
+                
+                asdf install golang 1.21.5
```
lpusok, line 196: Why a specific go version is required?

lpusok (approved):

bitrise.yml
zsolt-marta-bitrise: Older CLI versions used this one, as the project grows and (hopefully) we update the version used by it we'll need to install more older ones

Badlazzor (approved):

PR #115, opened 2025-04-23
feat: ACI-3637 Add timestamps flag for bazel and wait for BES upload

zsolt-marta-bitrise:


cmd/enableForBazel.go
```diff
@@ -49,9 +49,11 @@ If the "# [start/end] generated-by-bitrise-build-cache" block is already present
 }
 
 var rbeEnabled bool //nolint:gochecknoglobals
+var timestamps bool //nolint:gochecknoglobals
 
 func init() {
 	enableForBazelCmd.Flags().BoolVar(&rbeEnabled, "with-rbe", false, "Enable Remote Build Execution (RBE)")
+	enableForBazelCmd.Flags().BoolVar(&timestamps, "timestamps", false, "Enable timestamps in build output")
```
zsolt-marta-bitrise, line 56: Open for suggestions

internal/config/bazel/bazel_config.go
```diff
@@ -19,16 +19,24 @@ var (
...
+	RBEEndpointURL      string
+	WorkspaceID         string
+	AuthToken           string
+	IsTimestampsEnabled bool
 	// Metadata
 	CacheConfigMetadata common.CacheConfigMetadata
 }
 
+type Preferences struct {
+	RBEEndpointURL      string
```
zsolt-marta-bitrise, line 32: IMO RBE is a preference rather than a common config like the auth url, but can be challenged

cmd/enableForBazel.go
imrekel: fine with me...

internal/config/bazel/bazel_config.go
imrekel: what's the definition of preference vs config? :) but yep, I think I get it although I don't remember seeing this distinction often (at least in our codes)

zsolt-marta-bitrise: That's a good point, I just used what we have in Gradle, I normally don't use prefs other then user settings

PR #127, opened 2025-06-04
[ACI-3750] Add activate command

Badlazzor:


cmd/activateForGradle_test.go
```diff
@@ -0,0 +1,675 @@
...
+        // when
+        err := activateForGradleCmdFn(mockLogger, tmpGradleHomeDir, envVars)
+
+        // then
+        require.NoError(t, err)
+        //
+        initGradlecontent, err := ReadActivatedPluginsFromInitGradle(filepath.Join(tmpGradleHomeDir, "init.d", "bitrise-build-cache.init.gradle.kts"))
+        require.NoError(t, err)
+        expected := `import io.bitrise.gradle.cache.BitriseBuildCache
+import io.bitrise.gradle.cache.BitriseBuildCacheServiceFactory
```
zsolt-marta-bitrise, line 232: Do we need to test the output config file in a detailed way? We have tests for the config in `gradleconfig_test.go`

cmd/enableForGradle.go
```diff
@@ -72,7 +72,7 @@ func init() {
 	enableForGradleCmd.Flags().StringVar(&paramRemoteCacheEndpoint, "remote-cache-endpoint", "", "Remote cache endpoint URL")
 }
 
-func writeGradleInit(logger log.Logger, gradleHomePath string, endpointURL string, authToken string, cacheConfigMetadata common.CacheConfigMetadata, prefs gradleconfig.Preferences) error {
+func writeGradleInit(logger log.Logger, gradleHomePath string, prefs gradleconfig.Preferences) error {
```
zsolt-marta-bitrise, line 75: Are we keeping this for backwards-compatibility (and minor version reelease)?

cmd/activateForGradle_test.go
```diff
@@ -0,0 +1,675 @@
...
+
+initscript {
+    repositories {
+        mavenLocal()
+        mavenCentral()
+        google()
+        maven(url="https://jitpack.io")
+    }
+    dependencies {
+        classpath("io.bitrise.gradle:gradle-analytics:2.1.28")
```
zsolt-marta-bitrise, line 378: Yeah these will also have to be updated in the bump automation, I think it's not ideal

internal/config/gradle/gradleconfig.go
```diff
@@ -28,59 +29,131 @@ var (
...
-func GenerateInitGradle(endpointURL, authToken string, preferences Preferences, cacheConfigMetadata common.CacheConfigMetadata) (string, error) {
+func GenerateInitGradle(preferences Preferences) (string, error) {
 	// required check
-	if len(authToken) < 1 {
+	if len(preferences.AuthToken) < 1 {
 		return "", fmt.Errorf("generate init.gradle, error: %w", errAuthTokenNotProvided)
 	}
 
-	if len(endpointURL) < 1 {
+	if len(preferences.Cache.EndpointURL) < 1 && preferences.Cache.Usage == UsageLevelEnabled {
```
zsolt-marta-bitrise, line 113: We should add the same check for analytics too, right?

internal/config/gradle/gradleconfig.go
```diff
@@ -28,59 +29,131 @@ var (
...
+}
+
+type CacheTemplateInventory struct {
+	Version             string
+	Enabled             bool
+	Dependency          bool
+	EndpointURLWithPort string
+	IsPushEnabled       bool
+	ValidationLevel     string
+	Metadata            common.CacheConfigMetadata
```
zsolt-marta-bitrise, line 73: Should we move this to the top level? All plugins need the metadata

internal/config/gradle/initd.gradle.kts.gotemplate
```diff
@@ -9,54 +9,71 @@ initscript {
...
+            authToken.set("{{ .AuthToken }}")
+            dumpEventsToFiles.set({{ .IsDebugEnabled }})
+            debug.set({{ .IsDebugEnabled }})
 
-{{- if .IsAnalyticsEnabled }}
+            providerName.set("{{ .Cache.Metadata.CIProvider }}")
 
+            bitrise {
+                {{- if .AppSlug }}
+                appSlug.set("{{ .AppSlug }}")
```
zsolt-marta-bitrise, line 58: Why not in the metadata?

internal/config/gradle/gradleconfig.go
Badlazzor: Sure

internal/config/gradle/initd.gradle.kts.gotemplate
Badlazzor: Moving the metadata outside and making this part of it

cmd/activateForGradle_test.go
Badlazzor: The const strings will be used in tests so we only need to update them there.

internal/config/gradle/gradle_properties_from_params_test.go
```diff
@@ -0,0 +1,134 @@
...
+		tmpGradleHomeDir := filepath.Join(tmpPath, ".gradle")
+		_ = os.MkdirAll(tmpGradleHomeDir, 0755)
+
+		propertyFilePath := filepath.Join(tmpGradleHomeDir, "gradle.properties")
+
+		return mockLogger, tmpGradleHomeDir, propertyFilePath
+	}
+
+	t.Run("Update gradle properties", func(t *testing.T) {
+		updater := GradlePropertiesUpdater{DefaultOsProxy()}
```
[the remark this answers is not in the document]
zsolt-marta-bitrise: You probably wanted this in `prep()` but then it needs to be passed to the updater too?

cmd/activate_for_gradle.go
```diff
@@ -0,0 +1,113 @@
...
+	templateInventory, err := templateInventoryProvider(logger, envProvider, isDebugLogMode)
+	if err != nil {
+		return err
+	}
+
+	if err := templateWriter(
+		templateInventory,
+		logger,
+		gradleHomePath,
+		gradleconfig.DefaultOsProxy(),
```
zsolt-marta-bitrise, line 102: nit: this should be also DI'd like the others (e.g. envprovider), and it doesn't even need to be part of the templatewriter's signature as it should be there when the writer struct is created/mocked

zsolt-marta-bitrise (approved):
Nice work 🚀

PR #134, opened 2025-06-12
[ACI-3782] Fetch bitrise plugins to local repository

Badlazzor:
Fetch bitrise plugins to local repository when enabling gradle buildcache


cmd/enableForGradle.go
```diff
@@ -169,3 +176,40 @@ func enableForGradleCmdFn(logger log.Logger, gradleHomePath string, envProvider
...
+
+func getPlugins(ctx context.Context, logger log.Logger, envProvider func(string) string) error {
+	// Required configs
+	logger.Infof("(i) Check Auth Config")
+	authConfig, err := common.ReadAuthConfigFromEnvironments(envProvider)
+	if err != nil {
+		return fmt.Errorf("read auth config from environment variables: %w", err)
+	}
+
+	kvClient, err := createKVClient(ctx,
```
imrekel, line 188: an idea for the future: let's call it `createBuildCacheKVClient` to distinguish it from the actual Key Value Cache service - which stores data in cloud storage)

cmd/enableForGradle.go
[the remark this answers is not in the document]
Badlazzor: 👍  will rename them to snake case

internal/gradle/plugin_cacher.go
```diff
@@ -0,0 +1,127 @@
...
+		logger.Infof("(i) Fetching " + plugin.id + ":" + plugin.version)
+
+		go func(plugin Plugin) {
+			defer wg.Done()
+			defer func() { <-semaphore }() // Release a slot in the semaphore
+
+			for _, file := range plugin.files() {
+				// Try to fetch from cache
+				if downloaded, err := pluginCacher.fetchFromCache(ctx, kvClient, file); err == nil {
+					if !downloaded {
```
zsolt-marta-bitrise, line 46: Shouldn't this be in reverse?

internal/gradle/plugin_cacher.go
```diff
@@ -0,0 +1,127 @@
...
+					continue
+				}
+
+				// Try to download from artifact repositories
+				downloader := PluginDownloader{logger: logger}
+				source, err := downloader.Download(file)
+				if err != nil {
+					errs = append(errs, err)
+
+					return
```
zsolt-marta-bitrise, line 61: Not `continue`?

internal/gradle/plugin.go
```diff
@@ -0,0 +1,115 @@
...
+		groupPath,
+		gf.id,
+		gf.version,
+	)
+}
+
+func (gf *PluginFile) dir() string {
+	home, err := os.UserHomeDir()
+	if err != nil {
+		home = os.Getenv("PWD")
```
zsolt-marta-bitrise, line 53: I think we should log this as a warning/info if we fall back to the current dir

internal/gradle/plugin.go
```diff
@@ -0,0 +1,115 @@
...
+	return fmt.Sprintf(
+		"%s-%s%s.%s",
+		gf.id,
+		gf.version,
+		classifierPart,
+		gf.extension,
+	)
+}
+
+func (gf *PluginFile) path() string {
```
zsolt-marta-bitrise, line 39: nit: usually path contains both the parent dir's path and the name

cmd/enable_for_gradle.go
```diff
@@ -169,3 +176,40 @@ func enableForGradleCmdFn(logger log.Logger, gradleHomePath string, envProvider
...
+			Logger: logger,
+		})
+	if err != nil {
+		return fmt.Errorf("create kv client: %w", err)
+	}
+
+	pluginCacher := gradle.PluginCacher{}
+
+	if err = pluginCacher.CachePlugins(ctx, kvClient, logger, []gradle.Plugin{
+		gradle.PluginAnalytics(),
```
zsolt-marta-bitrise, line 208: and not rbe?

internal/gradle/plugin_cacher.go
zsolt-marta-bitrise: ah I see, that would work thanks

internal/gradle/plugin.go
```diff
@@ -0,0 +1,115 @@
...
+	if err != nil {
+		home = os.Getenv("PWD")
+	}
+
+	return filepath.Join(home, ".m2", "repository", gf.path())
+}
+
+func (gf *PluginFile) key() string {
+	return fmt.Sprintf(
+		"%s:%s-test",
```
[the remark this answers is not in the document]
Badlazzor: sorry was left there from testing the results should be without

internal/gradle/plugin.go
zsolt-marta-bitrise: either is fine by me 👍🏻

internal/gradle/plugin_cacher.go
zsolt-marta-bitrise: Ah ok, this returns from the goroutine only (I miss the Kotlin syntax :D)

internal/gradle/plugin_cacher.go
Badlazzor: Will be done in https://github.com/bitrise-io/bitrise-build-cache-cli/pull/127

internal/gradle/plugin.go
Badlazzor: Will be done in https://github.com/bitrise-io/bitrise-build-cache-cli/pull/127

PR #137, opened 2025-06-24
feat: ACI-3805 Don't restore config cache if transforms missing

zsolt-marta-bitrise:


zsolt-marta-bitrise:
Closing as transforms directory can even be present after sdk installation

PR #141, opened 2025-06-27
fix: ACI-3819 Send metadata for local

zsolt-marta-bitrise:


internal/config/bazel/activate_bazel_params.go
```diff
@@ -88,11 +88,7 @@ func (params ActivateBazelParams) commonTemplateInventory(
 	}
 
 	cacheConfig := common.NewMetadata(envProvider,
```
zsolt-marta-bitrise, line 90: Eventually, we should also DI the cache comfig provider, but I wanted to limit the scope here

internal/config/bazel/activate_bazel_params_test.go
```diff
@@ -52,6 +72,10 @@ func Test_ActivateBazelParams(t *testing.T) {
 		assert.Equal(t, "WorkflowName1", inventory.Common.WorkflowName)
 		assert.Equal(t, "BuildID1", inventory.Common.BuildID)
 		assert.True(t, inventory.Common.Timestamps)
+		assert.Equal(t, "bitrise", inventory.Common.CIProvider)
+
+		assert.Equal(t, "Linux", inventory.Common.HostMetadata.OS)
```
zsolt-marta-bitrise, line 77: These are needed because we don't DI the cache config provider

internal/config/common/cache_config.go
```diff
@@ -114,7 +114,9 @@ func NewMetadata(envProvider EnvProviderFunc, commandFunc CommandFunc, logger lo
 			bitriseCISpecificMetadata{}, hostMetadata)
 	}
 
-	return CacheConfigMetadata{}
+	return CacheConfigMetadata{
+		HostMetadata: hostMetadata,
```
zsolt-marta-bitrise, line 118: The real fix

PR #149, opened 2025-07-10
fix: Disable push if RBE enabled

zsolt-marta-bitrise:
For https://bitrise.atlassian.net/wiki/spaces/RD/pages/4057202703/Devcenter+article+Bitrise+Remote+Execution+for+Bazel#--noremote_upload_local_results


zsolt-marta-bitrise:
After discussion with @ramil-bitrise, adding this to the step level to allow overriding manually (locally) and to match BuildBuddy

PR #271, opened 2026-04-14
fix: Don't fail when CLI is the main binary with (devel) version

Badlazzor:
## Summary

When the CLI binary runs `activate react-native`, `cliVersion()` fails because `debug.ReadBuildInfo().Main.Version` is `(devel)` for binaries built by GoReleaser (which uses `go build`, not `go install module@version`).

Since `Main.Path` matches `cliModulePath`, we know we ARE the CLI binary — there's nothing to install. Return the version as-is and let `EnsureAll` skip via `IsInstalled()` (the binary is already on PATH).

**Error before fix:**
```
Error: activate react-native: install dependencies: determine CLI tool: determine CLI version: module github.com/bitrise-io/bitrise-build-cache-cli/v2 not found in build info (main=github.com/bitrise-io/bitrise-build-cache-cli/v2, version=(devel))
```

## Test plan

- [x] `make check` passes
- [ ] CI E2E workflows pass
- [ ] RN step E2E: `activate react-native` succeeds with CLI binary from GitHub releases

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #275, opened 2026-04-20
fix: Self-install CLI binary when not on PATH during activate

Badlazzor:
## Summary
- When CLI is the main binary but not on PATH (e.g. `go run` dev builds), `IsInstalled()` returns false and download is attempted with `(devel)` version → HTTP 404
- Fix: detect we ARE the CLI binary via `isMainCLIBinary()` and copy `os.Executable()` to `/usr/local/bin` instead of downloading from GitHub
- Completes the fix from #271 which handled `cliVersion()` but not the `exec.LookPath` failure case

## What changed
`CLITool()` now branches on `isMainCLIBinary()`:
- **true** (we are the CLI): Install = `selfInstall()` — copies running executable to `/usr/local/bin/bitrise-build-cache`
- **false** (step binary with CLI as dependency): Install = download from GitHub releases (unchanged)

## Test plan
- [ ] `go run . activate react-native` no longer fails with 404 on `v(devel)`
- [ ] Released step binary still downloads CLI from GitHub releases normally
- [ ] `make check` passes (lint + unit tests)

### DIFF
--- a/internal/dependencies/cli.go
+++ b/internal/dependencies/cli.go
@@ -24,23 +24,85 @@ const (
 
 // CLITool returns a Tool that installs the bitrise-build-cache binary
 // matching the version embedded in the current binary's module dependencies.
+// When the current process IS the CLI binary (e.g. dev builds via `go run`),
+// it self-installs by copying the running executable to InstallDir.
 func CLITool() (Tool, error) {
 	version, err := cliVersion()
 	if err != nil {
 		return Tool{}, fmt.Errorf("determine CLI version: %w", err)
 	}
 
+	install := func(ctx context.Context, logger log.Logger) error {
+		return installFromGitHubRelease(
+			ctx, logger,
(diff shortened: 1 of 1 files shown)


zsolt-marta-bitrise:
[//]: # ([skip ci])



## Summary
Added self-installation capability for the CLI binary when running as the main binary (e.g., `go run` dev builds). The CLI now detects if it's the main binary and copies itself to `/usr/local/bin/bitrise-build-cache` instead of attempting to download from GitHub releases.


## Walkthrough
| File | Summary |
|------|---------|
| internal/dependencies/cli.go | Added isMainCLIBinary() to detect if current process is the CLI binary, selfInstall() to copy running executable to /usr/local/bin, and modified CLITool() to branch on isMainCLIBinary() for self-install vs GitHub download |

zsolt-marta-bitrise (approved):
approving to unblock - we should remove gh install completely and always copy the binary

PR #277, opened 2026-04-21
feat(cli): add status command and React Native activation marker

Badlazzor:
## Summary

Adds `bitrise-build-cache status`, a query API for \"which build cache features are enabled on this machine?\" — intended to be consumed by steps that want to transparently wrap their build commands (e.g. `bitrise-build-cache react-native run --`) when the RN cache is active.

Also adds a dedicated React Native activation marker (`~/.bitrise/cache/reactnative/config.json`) written at the end of `react-native activate`. This decouples \"RN cache is active\" from the individual sub-feature (gradle / xcode / cpp) flags.

## What changed

- **`bitrise-build-cache status`** — new root command with four output modes:
  | Invocation | Behaviour |
  |---|---|
  | `status` | Text table of all features |
  | `status --json` | JSON object of all features |
  | `status --feature=<n>` | `enabled` / `disabled` |
  | `status --feature=<n> --quiet` | No stdout; exit 0 (enabled), 1 (disabled), 2 (unknown) |
  | `status --feature=<n> --json` | `{\"<n>\": bool}` |
- **`pkg/status`** — exported `Checker` (public API for Go consumers). Follows the established `exported struct, DI via params` pattern from `CLAUDE.md`.
- **`internal/config/reactnative`** — new marker file read/write; `Enabled=true` is the signal consumers key on.
- **`pkg/reactnative.Activator`** — writes the marker at the end of `Activate()`, recording which sub-features (Gradle/Xcode/Cpp) were activated and the CLI version.
- **`cmd/common.Execute`** — now translates a `statusExitError` into the appropriate `os.Exit` code so `--quiet` returns 1 without cobra printing a spurious error.

### Detection signals

| Feature | Signal |
|---|---|
| gradle | `~/.gradle/init.d/bitrise-build-cache.init.gradle.kts` exists |
| xcode | `~/.bitrise-xcelerate/config.json` with `buildCacheEnabled=true` |
| cpp | `~/.bitrise/cache/ccache/config.json` with `enabled=true` |
| react-native | new marker with `enabled=true` |
| bazel | stub `false` (TODO — activation detection deferred; no consumer yet) |

## Test plan

- [x] Unit tests pass (`go test -tags unit -race ./...`)
- [x] `make check` passes (go generate + lint-fix + unit tests)
- [x] Manual smoke test of binary:
  - `status` prints text table
  - `status --json` prints expected JSON shape
  - `status --feature=react-native --quiet` exits 1 when no marker is present
- [ ] On a CI machine with `react-native activate` run end-to-end, verify marker is written and `--feature=react-native --quiet` exits 0
- [ ] Follow-up: release CLI (per `.claude/skills/release/SKILL.md`) before downstream step changes (Android / iOS) can land — those PRs PATH-probe the released binary

## Follow-ups (not in this PR)

- Android step wraps gradle with `bitrise-build-cache react-native run --` when this command reports RN enabled.
- iOS step wraps xcodebuild the same way.
- Wire bazel detection once activation has a stable marker.

### DIFF
--- a/cmd/common/root.go
+++ b/cmd/common/root.go
@@ -26,6 +26,10 @@ In case of Bazel it's done via creating or modifying $HOME/.bazelrc.`,
 func Execute() {
 	err := RootCmd.Execute()
 	if err != nil {
+		if code, ok := HandleStatusExit(err); ok {
+			os.Exit(code)
+		}
+
 		os.Exit(1)
 	}
 }
--- a/cmd/common/status.go
+++ b/cmd/common/status.go
@@ -1,0 +1,165 @@
+package common
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"text/tabwriter"
+
+	"github.com/spf13/cobra"
+
+	"github.com/bitrise-io/bitrise-build-cache-cli/v2/pkg/status"
+)
+
+//nolint:gochecknoglobals
+var (
+	statusJSONOutput bool
+	statusFeature    string
+	statusQuiet      bool
+)
+
+//nolint:gochecknoglobals
+var statusCmd = &cobra.Command{
+	Use:           "status",
+	Short:         "Show which Bitrise Build Cache features are enabled on this machine",
+	Long:          "Reports gradle / xcode / cpp / react-native / bazel activation status. Intended for step integrations that need to decide whether to engage cache wrapping.",
+	SilenceUsage:  true,
+	SilenceErrors: true,
+	RunE: func(cmd *cobra.Command, _ []string) error {
+		return runStatus(cmd.OutOrStdout(), cmd.ErrOrStderr(), status.NewChecker(status.CheckerParams{}))
+	},
+}
+
+func init() {
+	statusCmd.Flags().BoolVar(&statusJSONOutput, "json", false, "Emit machine-readable JSON instead of a text table")
+	statusCmd.Flags().StringVar(&statusFeature, "feature", "", "Query a single feature: gradle, xcode, cpp, react-native, bazel")
+	statusCmd.Flags().BoolVar(&statusQuiet, "quiet", false, "Suppress stdout; only meaningful with --feature (exit 0=enabled, 1=disabled)")
+
+	RootCmd.AddCommand(statusCmd)
+}
+
+// statusExitError lets us signal a non-zero exit code without letting cobra
+// print the error (we want silent-exit semantics for `--quiet`).
+type statusExitError struct{ code int }
+
+func (e *statusExitError) Error() string { return fmt.Sprintf("status exit code %d", e.code) }
+
+// HandleStatusExit converts a statusExitError returned by Execute into an
+// os.Exit code. Other errors fall through to the caller.
+func HandleStatusExit(err error) (int, bool) {
+	var se *statusExitError
+	if errors.As(err, &se) {
+		return se.code, true
+	}
+
+	return 0, false
+}
+
+func runStatus(out, errOut io.Writer, checker *status.Checker) error {
+	if statusFeature != "" {
+		return runStatusFeature(out, errOut, checker)
+	}
+
+	if statusQuiet {
+		fmt.Fprintln(errOut, "error: --quiet requires --feature")
+
+		return &statusExitError{code: 2}
+	}
+
+	s := checker.Status()
+	if statusJSONOutput {
+		return writeJSON(out, s)
+	}
+
+	return writeTable(out, s)
+}
+
+func runStatusFeature(out, errOut io.Writer, checker *status.Checker) error {
+	enabled, err := checker.IsEnabled(statusFeature)
+	if err != nil {
+		if errors.Is(err, status.ErrUnknownFeature) {
+			fmt.Fprintf(errOut, "error: unknown feature %q (expected: gradle, xcode, cpp, react-native, bazel)\n", statusFeature)
+
+			return &statusExitError{code: 2}
+		}
+
+		return fmt.Errorf("query status: %w", err)
+	}
+
+	if statusQuiet {
+		if enabled {
+			return nil
+		}
+
+		return &statusExitError{code: 1}
+	}
+
+	if statusJSONOutput {
+		payload := map[string]bool{jsonKey(statusFeature): enabled}
+
+		return writeJSON(out, payload)
+	}
+
+	if enabled {
+		fmt.Fprintln(out, "enabled")
+	} else {
+		fmt.Fprintln(out, "disabled")
+	}
+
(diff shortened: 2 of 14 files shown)


zsolt-marta-bitrise:
[//]: # ([skip ci])



## Summary
Adds `bitrise-build-cache status` command to query which build cache features are enabled, with text/JSON output modes and quiet exit codes. Introduces React Native activation marker (`~/.bitrise/cache/reactnative/config.json`) written at end of `react-native activate`. Exports public `Checker` API in `pkg/status` for Go consumers.


## Walkthrough
| File | Summary |
|------|---------|
| cmd/common/root.go, cmd/common/status.go | Added status command with four output modes: text table, JSON, single feature query, and quiet mode with exit codes. Added HandleStatusExit to translate statusExitError into appropriate os.Exit codes. |
| pkg/status/checker.go | New public API struct Checker with Status() and IsEnabled() methods to query feature enablement. Detects gradle, xcode, cpp, and react-native via config files. |
| pkg/reactnative/activator.go | Added saveReactNativeMarker() method to write activation marker at end of Activate() flow, recording enabled status and CLI version. |
| internal/config/reactnative/config.go | New marker file implementation with Read/Write functions for JSON config containing Enabled boolean flag, with directory fallback logic. |
| internal/config/xcelerate/config.go | Improved dependency injection by replacing direct os.OpenFile() call with osProxy parameter in ReadConfig(). |

viktorbenei (commented):
**Overall**: Solid, well-scoped PR. Design is clean, tests are comprehensive, signals match the activation side. Reads like ~90% done. A few items worth addressing before merge, grouped by severity.

---

### 🔶 Design — worth pushing back on

**1. `bazel` stub lies to callers.** `bazelEnabled()` returns `false` unconditionally, so on any machine where bazel cache IS genuinely active, `status --feature=bazel --quiet` exits `1` (disabled) rather than `2` (unknown). A shell integration that trusts the contract will wrap the wrong binary. The TODO comment is honest, but the behavior isn't. Two cleaner options:
- Drop bazel from `Status` + `IsEnabled` until there's a detection signal (simplest).
- Return `ErrUnknownFeature` for bazel so it exits 2 and is clearly gated from the set.

The current halfway choice ("list it, but always say no") is the only bad option.

**2. Speculative fields in the RN marker.** `rnconfig.Config` stores `Version`, `ActivatedAt`, `Gradle`, `Xcode`, `Cpp`, `AuthConfig` — none of which any consumer in this PR reads. `Checker.reactNativeEnabled()` only checks `Enabled`. I'd trim to `Enabled` + whatever a named follow-up actually needs. Shipping the wider schema now locks you into its JSON shape before anyone's validated it against a real consumer.

Same critique applies to the exported `rnconfig.Remove` — zero call sites, only self-tested. Either wire a `react-native deactivate` that calls it (worth doing anyway, for symmetry with activate) or drop it.

**3. AuthConfig in the marker file.** The marker is written via `os.Create` (mode 0644 after umask) and contains `CacheAuthConfig` — i.e., the auth token is now in yet another on-disk location. This is consistent with `ccache` and `xcelerate` configs (pre-existing pattern), so it's not a regression. But those files at least have a runtime reason to hold the token. Here the marker's sole documented purpose is `status` (doesn't need auth) and the `react-native run --` wrapper that hasn't been written yet. If that wrapper can read auth from env or from the multiplatform config, I'd drop `AuthConfig` from the marker and shrink blast radius. What concrete thing needs AuthConfig in this specific file?

---

### 🟡 Correctness & consistency

**4. `AuthConfig` json tag missing `omitempty`** — `ccache` and `xcelerate` both use `json:\"authConfig,omitempty\"`; rnconfig has `json:\"authConfig\"`. Cosmetic, but if you keep AuthConfig, match the pattern.

**5. `rnconfig.DirPath` omits the `Executable()` fallback** that `xcelerate.DirPath` and `ccache.DirPath` include. Probably fine (same final fallback to `.`), but inconsistent — pick one convention.

**6. `Checker.xcodeEnabled()` double-stubs the osProxy.** `xcelerate.ReadConfig` uses `os.OpenFile` directly (pre-existing bug in that file, line 68), not `osProxy.OpenFile`. In the status tests this works only because the fixture is written to a real temp dir that `os.OpenFile` can then read. Not introduced by this PR, but now a second code path depends on that inconsistency. Worth a follow-up to fix `xcelerate/config.go`.

**7. `--json` + `--quiet` + `--feature=…` silently drops `--json`.** In `runStatusFeature`, the `statusQuiet` branch returns before the `statusJSONOutput` branch runs. Either error on the combination or document that `--quiet` wins.

**8. `--quiet` flag help text says `exit 0=enabled, 1=disabled`** but exit 2 (unknown feature / misuse) is also possible. One-liner doc fix.

---

### 🟢 Test hygiene (nits, not blockers)

**9. `TestStatus_TextTable` doesn't actually verify correctness.** It asserts each label (\"gradle\", \"xcode\", …, \"enabled\", \"disabled\") appears *somewhere* in stdout. It does NOT assert that `xcode` maps to `enabled` and `gradle` to `disabled` — a bug that swapped row order or labels would pass. Tighten with full-string match or per-row `assert.Regexp`.

**10. Cobra flag state is reset in `t.Cleanup`** — works today because tests aren't parallel and don't panic before registration, but fragile. Safer to reset as the first act of `runStatusCmd` (explicit `cmd.Flags().Set(...)` before `Execute`). Also `RootCmd.SetOut/SetErr` are never restored.

**11. `TestNewChecker_Defaults` hits the real `\$HOME`.** Documented as such, but asserts nothing beyond \"doesn't panic\". Low value, harmless.

**12. No integration-level test that `Execute()` actually translates `statusExitError` into the right `os.Exit` code.** The PR description's test plan flags this as manual-only. Given the whole point of the custom error type is the exit code, a subprocess test (\`cmd.Test\`-style) would be worth the 20 lines.

---

### ✅ Things I liked

- Detection signals deliberately mirror what activate writes (no separate state store) — exactly right.
- `SilenceUsage: true, SilenceErrors: true` with explicit error writes + typed-error-driven exit code — clean.
- `CheckerParams` with nil→default fallback in `NewChecker` follows the DI pattern.
- Exit-code contract documented in a table in the PR description.
- The `jsonKey` mapping to keep `--feature=react-native --json` shape-compatible with `--json` is a nice touch.
- Marker write is the last step of `Activate()`, so a partial activation doesn't leave a truthy marker. Good sequencing.

---

### Suggested before merge

Minimum I'd ask for: **(1)** decide on bazel — remove or make unknown, **(3)** justify or drop AuthConfig in the marker, **(9)** tighten the table test to actually validate row→label mapping. Everything else can be follow-ups.

zsolt-marta-bitrise (approved):

PR #282, opened 2026-04-23
feat: aggregate child invocation hit rates into parent invocation

Badlazzor:
## Summary
- New `pkg/common/childstats` package: local file ledger at `~/.bitrise/cache/invocations/<parent-id>/<child-id>.json` with atomic writer, reader/aggregator, 7-day TTL sweep.
- Ccache storage helper and xcode wrapper now write ledger entries when a parent invocation ID is set.
- React-native wrapper aggregates child entries (unweighted mean, baseline filtered) and reports the result via a new `multiplatform.Invocation.HitRate` field (`json:"hitRate"`, mirroring the xcelerate shape).
- New `report-child-stats` subcommand for external callers (gradle plugin writes directly, no subprocess).
- `activate` subcommand runs an opportunistic sweep via `PersistentPreRun`.

## Why
Product wants the react-native wrapper invocation to report one `hitRate` number that averages its children's hit rates (gradle, xcode, ccache). No backend aggregation today; backend already accepts the field on the existing invocation endpoint, so this is a one-repo PR from the API's perspective.

Companion PR on `bitrise-io/gradle-plugins`: https://github.com/bitrise-io/gradle-plugins/pull/new/feat/child-invocation-hit-rate-ledger — adds the gradle-side writer on the same path contract.

## Test plan
- [x] `go test -tags unit -race ./...` — green
- [x] `make lint` — green
- [ ] End-to-end: run an RN build on CI with the gradle plugin companion PR and verify `hitRate` appears on the wrapper invocation payload + log line `Cache hit rate: X.X% (avg of N child invocations)`
- [ ] Verify baseline benchmark runs do not pollute the mean (baseline entries excluded)
- [ ] Verify `~/.bitrise/cache/invocations/` gets swept on next `activate` after a 7-day-old orphan

## Notes
- CLI-only changes; schema already on the backend (mirrors xcelerate `hitRate`).
- Plugin-side change is separate and can ship independently — older plugin versions just don't contribute to the mean.
- Ccache ledger uses `configcommon.BenchmarkPhaseEnvVar("ccache")` so if per-tool ccache phase is ever exported, baselines are filtered correctly.


cmd/common/report_child_stats.go
```diff
@@ -0,0 +1,67 @@
...
+		if err := childstats.NewWriter().Write(entry); err != nil {
+			return fmt.Errorf("write child stats ledger: %w", err)
+		}
+
+		return nil
+	},
+}
+
+func init() {
+	reportChildStatsCmd.Flags().StringVar(&reportChildParentID, "parent-id", "", "Parent invocation ID (required; no-op if empty)")
```
zsolt-marta-bitrise: nit: Call these `reportParentID` and `reportChildID`, having childchild in the name is confusing

cmd/xcode/xcodebuild.go
```diff
@@ -357,6 +358,24 @@ func (c *XcodebuildRunner) saveInvocationAndRelation(inv analytics.Invocation) {
...
 	if parentID := os.Getenv("BITRISE_INVOCATION_ID"); parentID != "" {
 		c.sendRelation(parentID)
+		c.writeChildStatsLedger(parentID, inv)
+	}
+}
+
+// writeChildStatsLedger records this xcode invocation's hit rate in the
+// parent's local ledger so a parent wrapper (e.g. react-native) can
+// aggregate child hit rates at the end of its run.
+func (c *XcodebuildRunner) writeChildStatsLedger(parentID string, inv analytics.Invocation) {
```
zsolt-marta-bitrise: Can we save the xcod-reported targets as hits/total? That would make it closer to the real hit rate

cmd/common/report_child_stats.go
```diff
@@ -0,0 +1,67 @@
...
+	reportChildBuildTool      string
+	reportChildHitRate        float32
+	reportChildHits           int64
+	reportChildTotal          int64
+	reportChildBenchmarkPhase string
+)
+
+//nolint:gochecknoglobals
+var reportChildStatsCmd = &cobra.Command{
+	Use:          "report-child-stats",
```
zsolt-marta-bitrise: Is this used?

pkg/common/childstats/childstats.go
```diff
@@ -0,0 +1,293 @@
...
+		if entry.BenchmarkPhase == BenchmarkPhaseBaseline {
+			summary.SkippedCount++
+
+			continue
+		}
+
+		sum += entry.HitRate
+		summary.ChildCount++
+
+		byToolSums[entry.BuildTool] += entry.HitRate
```
zsolt-marta-bitrise, line 269: If we save the total/hit we could calculate based on that. I'd print in the logs both methods so that users can decide which one they believe (simple % avg, or total ratio of hits/total)

pkg/common/childstats/childstats.go
```diff
@@ -0,0 +1,293 @@
...
+
+		entry, err := readEntry(filepath.Join(dir, de.Name()))
+		if err != nil {
+			summary.SkippedCount++
+
+			continue
+		}
+
+		if entry.BenchmarkPhase == BenchmarkPhaseBaseline {
+			summary.SkippedCount++
```
zsolt-marta-bitrise: We still report hit rate for baseline builds in all build tools, it's OK to have 0% there, just add a log

pkg/common/childstats/childstats.go
```diff
@@ -0,0 +1,293 @@
...
+	byToolSums := map[string]float32{}
+	byToolCounts := map[string]int{}
+
+	for _, de := range entries {
+		if de.IsDir() || !strings.HasSuffix(de.Name(), ".json") {
+			continue
+		}
+
+		entry, err := readEntry(filepath.Join(dir, de.Name()))
+		if err != nil {
```
zsolt-marta-bitrise, line 243: Let's warn log

pkg/common/childstats/childstats.go
```diff
@@ -0,0 +1,293 @@
...
+}
+
+// LedgerDir returns the directory for a parent invocation's child entries.
+func LedgerDir(parentID string) string {
+	return filepath.Join(ledgerRoot(), parentID)
+}
+
+// LedgerPath returns the path of a single child entry file.
+func LedgerPath(parentID, childID string) string {
+	return filepath.Join(LedgerDir(parentID), childID+".json")
```
zsolt-marta-bitrise: Since the ledger root is commonly named (not specific to these stats), I'd add a common file suffix so we can later extend with other statistics shared among build tools. Or alternatively, use a subdir

zsolt-marta-bitrise: (this affects the gradle plugin too)

Badlazzor:
@zsolt-marta-bitrise addressed your feedback in 898c33c — re-review please:

- **Unused `report-child-stats` cmd** → deleted. No callers in this repo or gradle-plugins.
- **Xcode hits/total** → ledger entry now carries `runStats.CacheStats.Hits` / `TotalTasks`.
- **Aggregator: simple % avg vs total ratio** → both computed (`MeanHitRate` + `WeightedHitRate` / `TotalHits` / `TotalCount`, also per-tool). RN post-run logs both views.
- **Baseline entries** → no longer skipped; included with reported hit rate. New `Summary.BaselineCount`; warn log if baseline count > 0 so users see the drag.
- **Malformed entries** → warn log via new `Aggregator.Logger` field with file path + error.
- **Common file suffix** → entries now `<child-id>.childstats.json` (constant `EntryFileSuffix`). Plain `.json` files in the parent dir are ignored. ⚠️ Gradle plugin needs a matching change — separate PR coming.

zsolt-marta-bitrise (approved):

PR #296, opened 2026-04-29
feat: allowlist supported datacenters for gradle-mirrors activation

zsolt-marta-bitrise:
## Summary
- Only `IAD*`/`ORD*`/`AMS*` datacenters have a Bitrise mirror deployment, so restrict mirror activation to those.
- Other datacenters (e.g. `ATL1`) now log an info message and skip activation gracefully instead of writing an init script that points at a non-existent mirror host.
- Added `SupportedRegions` / `IsSupportedRegion` helper next to `DatacenterToRegion` and a corresponding skip branch in `Activate`.

## Test plan
- [x] `go test -tags unit -race ./internal/config/gradle/mirrors/...`
- [x] `make test-unit`
- [x] `make lint`
- [x] New `TestActivate` case for unsupported DC (`ATL1`) asserts no init script is written
- [x] New `TestIsSupportedRegion` covers iad/ord/ams/atl/empty

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #298, opened 2026-04-29
feat: robolectric repo mirror + gate gradle-plugins URL on supported regions

benbitrise:
## Summary

Two related changes to the gradle-mirrors path:

1. **Robolectric installation through the Bitrise Maven Central mirror.** Robolectric resolves its Android SDK jars at test time via the `robolectric.dependency.repo.url` system property; without overriding that, the Test task hits Maven Central directly even when the rest of the build is mirrored. Added a `UseAsRobolectricRepo bool` flag on `RepoMirror` and an `afterProject` block in the init template that sets `robolectric.dependency.repo.url` on every `Test` task to the matching mirror URL. The `mavencentral` mirror opts in via the new flag — previous draft of this used `{{ if eq .ID "Central" }}` partial matching on the template ID, replaced with the explicit boolean (mirroring the existing `ApplyToPluginManagement` pattern).

2. **Gate `GradlePluginsMirrorURL` on the supported-region allowlist.** `internal/config/gradle/mirror_url.go` was building `repository-manager-<region>...` URLs from any datacenter, bypassing the allowlist that `internal/config/gradle/mirrors.Activate` already honors. On customer-private GCP environments (e.g. `US_EAST4`) this emitted `repository-manager-us_east.services.bitrise.io:8090` into the build-cache init script's `pluginManagement.maven { url = uri(...) }` block, and the customer's GCP network couldn't reach that host — `Connect timed out`. The CLI v2.4.1 allowlist (PR #296) correctly skipped the standalone `activate-gradle-mirrors` step but didn't cover this code path. Now both paths share `mirrors.DatacenterToRegion` + `mirrors.IsSupportedRegion`, so unsupported DCs cleanly produce no mirror URL.

## Test plan

- [x] `make test-unit`
- [x] `make lint`
- [x] New `TestGradlePluginsMirrorURL` covers `AMS1` / `IAD1` (mirror URL emitted), `ATL1` and `US_EAST4` (empty), disabled, and missing-DC.
- [x] Existing `mirrors.TestActivate` cases still pass — robolectric block now keys off `UseAsRobolectricRepo` and is generated only for `mavencentral`.

### DIFF
--- a/internal/config/gradle/mirror_url.go
+++ b/internal/config/gradle/mirror_url.go
@@ -2,8 +2,8 @@ package gradleconfig
 
 import (
 	"fmt"
-	"strings"
-	"unicode"
+
+	"github.com/bitrise-io/bitrise-build-cache-cli/v2/internal/config/gradle/mirrors"
 )
 
 const (
@@ -18,7 +18,9 @@ const (
 
 // GradlePluginsMirrorURL returns the Bitrise-hosted gradle-plugins mirror URL
 // (proxying https://plugins.gradle.org/m2/) for the current datacenter, or ""
-// when the mirror is disabled or no datacenter is set (e.g. local dev).
+// when the mirror is disabled, no datacenter is set (e.g. local dev), or the
+// datacenter is outside the regions where a Bitrise mirror is deployed
+// (e.g. customer-private GCP envs like US_EAST4).
 func GradlePluginsMirrorURL(envs map[string]string) string {
 	if envs[MirrorEnabledEnvKey] != "true" {
 		return ""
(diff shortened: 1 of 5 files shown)


zsolt-marta-bitrise:
[//]: # ([skip ci])



## Summary
Adds support for Robolectric installation by exposing Maven Central mirror URL to test tasks. Refactors mirror URL logic to use helper functions, adds UseAsRobolectricRepo flag to mirror definitions, and extends the Gradle init script to configure robolectric.dependency.repo.url for applicable mirrors.


## Walkthrough
| File | Summary |
|------|---------|
| internal/config/gradle/mirror_url.go, mirror_url_test.go | Refactored to use DatacenterToRegion() and IsSupportedRegion() helpers; added check to return empty string for unsupported regions |
| internal/config/gradle/mirrors/mirror.go | Added UseAsRobolectricRepo boolean field to RepoMirror struct; set to true for mavencentral mirror |
| internal/config/gradle/mirrors/activate.go | Updated to pass UseAsRobolectricRepo field from mirror definition to template entry |
| internal/config/gradle/mirrors/asset/gradle-mirrors.init.gradle.kts.gotemplate | Extended Gradle init script to generate gradle.afterProject blocks that set robolectric.dependency.repo.url system property on Test tasks for mirrors with UseAsRobolectricRepo: true |

zsolt-marta-bitrise (approved):

PR #301, opened 2026-04-30
ci: also generate verification-metadata.xml with the Maven mirror enabled

zsolt-marta-bitrise:
## Summary

- Extend the `generate_gradle_verification_reference` step bundle to run `--write-verification-metadata sha256 help` a second time after activating the Bitrise Maven mirror, producing `cmd/gradle/asset/verification-metadata-mirror.xml` alongside the existing origin-repo asset.
- Widen `check_gradle_verification_reference` to assert both files are committed (switched to `git status --porcelain` so newly created files are also caught).
- Add the mirror metadata path to the `deploy-to-bitrise-io` artifact uploads in the `release` and `test` workflows.
- Seed `cmd/gradle/asset/verification-metadata-mirror.xml` with the empty Gradle dependency-verification scaffold so the file exists in HEAD; CI overwrites it with the real content.

## Why

Mirror-served artifacts can differ in bytes from the origin (e.g. POM / `.module` files generated by the proxy), so verification metadata recorded against the origin repos fails the build the moment the mirror is activated. We hit this in the DuckDuckGo workflow `test-dependency-verification-mirror`. Capturing a separate reference for the mirror lets the CLI later select the correct embed at runtime depending on whether the mirror is active.

## Test plan

- [ ] CI `test` workflow runs the bundle and produces the new asset; first run is expected to fail the check step asking for a manual commit of the regenerated `verification-metadata-mirror.xml` — once committed, subsequent runs are clean.
- [ ] No Go-side changes; `make lint` and `make test-unit` pass locally.
- [ ] `release` workflow uploads both verification-metadata XMLs as build artifacts.


zsolt-marta-bitrise:
[//]: # ([skip ci])



## Summary
Extended Gradle verification metadata generation to support the Bitrise Maven mirror. Added a second metadata generation run after activating the mirror, producing a separate reference file. Updated CI workflows to validate and deploy both verification metadata files.


## Walkthrough
| File | Summary |
|------|---------|
| bitrise.yml | Extended generate_gradle_verification_reference step to run twice (origin + mirror), updated check_gradle_verification_reference to validate both files, added mirror metadata deployment to release and test workflows |
| cmd/gradle/asset/verification-metadata-mirror.xml | Added seed file with empty Gradle dependency-verification scaffold for mirror metadata |

Badlazzor (approved):

PR #304, opened 2026-04-30
ci: rewrite dependency matrix to iterate CLI releases

zsolt-marta-bitrise:
## Summary

The dependency matrix used to walk step.yml versions in `bitrise-steplib/steps/activate-build-cache-for-gradle` and resolve each step version back to the CLI it shipped. That worked when only one step bundled the CLI, but the CLI is now embedded in several steps (Activate Gradle Remote Cache, Activate Gradle Mirrors, Activate React Native Features, Activate Gradle Features, Install Missing Android Tools, Gradle Runner) and they can ship different CLI versions independently. A step-indexed matrix only covers one installer and silently misses the rest.

Pivot the matrix to be CLI-release-indexed:

- For each stable `vMAJOR.MINOR.PATCH` CLI tag, download the `linux_amd64` release binary, run `activate gradle --cache --test-distribution -d` (or `enable-for gradle` on pre-`0.16.10` releases), and parse the generated init script for the Analytics / Remote Cache / Test Distribution plugin versions.
- Emit one row per CLI release with the release date.

Also drops the steplib clone, the gradle-step clone, and the asdf go installs from the workflow — we now run release binaries instead of `go run main.go`.

## Header

Rewritten to list all six consumer step repos and their links so users can find their installer regardless of which step the workflow uses, plus a callout that the matching CLI release page also has `verification-metadata.xml` (and `verification-metadata-mirror.xml` from v2.4.6+) attached for projects using Gradle dependency verification.

## Test plan

- [x] Smoke-tested locally on `v2.4.7`, `v2.4.6`, `v2.0.0`, and `v0.16.10` — all produce the expected init script and parse cleanly.
- [ ] CI run regenerates `docs/dependency-matrix.md` on `gh-pages` and the rendered page lists every CLI release with non-empty plugin columns.

### DIFF
--- a/scripts/generate-dependency-matrix.sh
+++ b/scripts/generate-dependency-matrix.sh
@@ -6,86 +6,86 @@ mkdir -p docs
 export RESULT_MD_PATH="docs/dependency-matrix.md"
 export MD_HEADER_PATH="assets/dependency-matrix-header.md"
 export CLI_RELEASE_URL_PREFIX="https://github.com/bitrise-io/bitrise-build-cache-cli/releases/tag"
+export CLI_DOWNLOAD_URL_PREFIX="https://github.com/bitrise-io/bitrise-build-cache-cli/releases/download"
 
 export BITRISE_BUILD_CACHE_WORKSPACE_ID=322a005426441b60
+# `activate gradle --test-distribution` requires this; on Bitrise CI it is set automatically.
+export BITRISE_APP_SLUG="${BITRISE_APP_SLUG:-dependency-matrix}"
 
 cat $MD_HEADER_PATH > $RESULT_MD_PATH
 export tmpdir=$(mktemp -d)
 
 pushd $tmpdir
 
-git clone https://github.com/bitrise-io/bitrise-steplib.git
-git clone https://github.com/bitrise-io/bitrise-build-cache-cli.git
-git clone https://github.com/bitrise-steplib/bitrise-step-activate-gradle-remote-cache.git
+git clone --filter=blob:none --no-checkout https://github.com/bitrise-io/bitrise-build-cache-cli.git
+cd bitrise-build-cache-cli
+git fetch --tags --no-recurse-submodules origin
+cd ..
 
 export tmp_md="release-table.md"
-echo "| Step version | CLI version | Analytics plugin version | Cache plugin version | Test Distribution plugin version |" >> $tmp_md
-echo "|--------------|-------------|--------------------------|----------------------|----------------------------------|" >> $tmp_md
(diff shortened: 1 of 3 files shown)


zsolt-marta-bitrise:
[//]: # ([skip ci])



## Summary
Rewrote the dependency matrix generation from step-indexed to CLI-release-indexed. The matrix now iterates over stable CLI releases, downloads pre-built binaries, runs activation commands, and parses plugin versions. Updated header documentation with consumer step links and Gradle dependency verification guidance. Removed unnecessary Go installation from workflow.


## Walkthrough
| File | Summary |
|------|---------|
| scripts/generate-dependency-matrix.sh | Pivoted matrix to iterate CLI releases instead of step versions. Downloads release binaries, runs activation commands, parses init scripts for plugin versions, and retrieves release dates from git tags. |
| assets/dependency-matrix-header.md | Updated title and added "Where the CLI runs" section listing six consumer steps with links. Expanded components section, added Gradle dependency verification guidance with verification-metadata files info. |
| bitrise.yml | Removed "Install go versions" step from workflow as pre-built CLI binaries no longer require Go installation. |

scripts/generate-dependency-matrix.sh
```diff
@@ -6,86 +6,86 @@ mkdir -p docs
...
-  semver_regex='v?([0-9]+)\.([0-9]+)\.([0-9]+)(-([0-9A-Za-z.-]+))?(\+([0-9A-Za-z.-]+))?'
-
-  if [[ $cli_version =~ $semver_regex ]]; then
-    major="${BASH_REMATCH[1]}"
-    minor="${BASH_REMATCH[2]}"
-    patch="${BASH_REMATCH[3]}"
-  else
-    echo "No semantic version found"
-    exit 1
+  tar -xzf "$cli_dir/cli.tar.gz" -C "$cli_dir"
```
zsolt-marta-bitrise, line 53: [bitrise-plugin-ai-reviewer]: scripts/generate-dependency-matrix.sh:53:0:e202d986

[//]: # ([skip ci])


**🐛 Bug**

The tar extraction command on line 53 is not protected by error handling. If the tar file is corrupted or the extraction fails, the script will exit due to `set -e` on line 3. This could cause the entire matrix generation to fail even if most CLI versions were processed successfully. The subsequent check on line 55 only verifies that the binary exists, but doesn't catch tar extraction errors.

<details>

```
In scripts/generate-dependency-matrix.sh at line 53, Add error handling to the tar extraction command on line 53.\nIf the extraction fails, log a warning and skip the current\nCLI version instead of exiting the entire script.
```

</details>

zsolt-marta-bitrise (commented):

PR #306, opened 2026-04-30
feat: disable jitpack mirror (returning 401)

Badlazzor:
## Summary

- Drop the `jitpack` entry from `KnownMirrors` in `internal/config/gradle/mirrors/mirror.go`. The cobra `--jitpack` flag is auto-derived from `KnownMirrors` and disappears with it.
- Update `TestActivate` AMS1-all-mirrors case to assert jitpack URL/predicate is NOT in the generated init script.

## Why

Bitrise's jitpack mirror endpoint is returning HTTP 401, breaking customer builds that route through it. Pulling the entry causes Gradle to fall back to direct `https://jitpack.io` resolution while the mirror is unavailable.

This is a temporary disable — the entry can be restored once the upstream auth issue is fixed.

## Test plan

- [x] `make lint` — 0 issues
- [x] `go test -tags unit -race ./internal/config/gradle/mirrors/... ./pkg/gradle/mirrors/...` — pass
- [ ] Manual: confirm generated `bitrise-gradle-mirrors.init.gradle.kts` no longer contains the jitpack predicate when running with all mirrors enabled.

### DIFF
--- a/internal/config/gradle/mirrors/activate_test.go
+++ b/internal/config/gradle/mirrors/activate_test.go
@@ -65,11 +65,12 @@ func TestActivate(t *testing.T) {
 			expectContains: []string{
 				"https://repository-manager-ams.services.bitrise.io:8090/maven/central",
 				"https://repository-manager-ams.services.bitrise.io:8090/maven/google",
-				"https://repository-manager-ams.services.bitrise.io:8090/maven/jitpack",
-				`"https://jitpack.io"`,
-				`"https://www.jitpack.io"`,
 				"https://repository-manager-ams.services.bitrise.io:8090/maven/gradle-plugins",
 				`r.getUrl().toString().trimEnd('/').equals("https://plugins.gradle.org/m2")`,
+			},
+			expectNotContain: []string{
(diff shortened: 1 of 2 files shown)


zsolt-marta-bitrise:
[//]: # ([skip ci])



## Summary
Disables the Bitrise jitpack mirror endpoint by removing the jitpack entry from KnownMirrors. The --jitpack cobra flag is auto-derived and disappears with it. Updated tests to verify jitpack mirror URLs are not in generated init scripts. Gradle falls back to direct jitpack.io resolution.


## Walkthrough
| File | Summary |
|------|---------|
| internal/config/gradle/mirrors/mirror.go, activate_test.go | Removed jitpack entry from KnownMirrors registry and updated TestActivate to assert jitpack URLs/predicates are not in generated init script |

Badlazzor:
Shelving — root cause is a different change, not the jitpack mirror entry itself. Reopening this would still be the right shape if we ever need to disable jitpack independently.

zsolt-marta-bitrise (approved):

PR #317, opened 2026-05-08
test(ACI-4754): assert RN invocations land on Monolith via BE queries

Badlazzor:
## Summary

[ACI-4754](https://bitrise.atlassian.net/browse/ACI-4754): close the loop on the RN e2e tests by querying the Monolith `BuildCache::InvocationsController` after each run and verifying the invocation actually landed on the BE — not just that the CLI claims it sent something.

Until now `scripts/react_e2e_assert_analytics.sh` only validated the CLI's own log output. If the BE silently dropped the invocation (auth issue, schema mismatch, sink outage), the e2e would still pass. This PR adds a complementary BE-side check.

## Changes

- `scripts/react_e2e_assert_invocations_be.sh` (new):
  - Extracts RN invocation ID from `$RN_CLI_LOG` and optional ccache / xcode child IDs from the CLI log + `xcelerate-*.log` files.
  - Polls `GET /build-cache/<workspace>/invocations/react-native/<id>.json` with retry (defaults: 6 attempts × 5 s), to ride out the BE's eventual-consistency window.
  - Asserts `id`, `buildTool`, `command`, `status`, `workflow`, `projectSlug` and cache stats (`cacheHits` / `cacheMisses`) on the parent record.
  - Queries `.../child-invocations.json` and asserts each ccache / xcode child detected locally is attached to the parent on BE.
  - Skips the children check when neither ccache nor xcode was active in the run.
- `bitrise_rn_config/bitrise.yml`: invokes the new script at the end of both `react-seek-android-e2e` and `react-seek-ios-e2e` workflows, after the existing log-based analytics assert.

Uses the existing app-level `MONOLITH_API_PAT` + `WORKSPACE_SLUG` envs (`bitrise.yml:16–19`) — they were prepared for this work.

## Knowns / risks

- **PAT aliasing.** `bitrise.yml` aliases `MONOLITH_API_PAT: $BITRISE_BUILD_CACHE_AUTH_TOKEN`. The Monolith controller uses `authenticate_user_from_personal_access_token!`, which expects a real Bitrise PAT, not the workspace cache auth token. If the workspace value isn't actually a PAT in the e2e env, the script will 401 against BE. First run will tell us; if so, we swap the env mapping to a dedicated PAT secret rather than fixing it in this script.
- **Field name guesses.** `cacheHits` / `cacheMisses` / `projectSlug` etc. are best-effort matches against `BuildToolInvocationInfoPresenter#to_h` — the script falls back to `.stats.cacheHits` and prints whatever it found. Easy follow-up to tighten once we have a live response sample.
- The BE check runs *after* the existing log-based assertion — log-side regressions still fail loud at the same point as before.

## Test plan

- [ ] CI green on this PR — main e2e workflows are manually triggered, but the `test` workflow lints/tests.
- [ ] Trigger `react-seek-android-e2e` manually after merge, confirm "Assert invocations are visible on Monolith BE" step passes against a real run.
- [ ] Trigger `react-seek-ios-e2e` manually, same.
- [ ] If BE fields differ from the script's guesses, follow up with a small adjustment PR (no need to gate this PR on it — failures are loud).

[ACI-4754]: https://bitrise.atlassian.net/browse/ACI-4754?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


zsolt-marta-bitrise:
[//]: # ([skip ci])

This PR is under review at commit `6350d01`. Please do not merge yet.

scripts/react_e2e_assert_invocations_be.sh
```diff
@@ -0,0 +1,191 @@
...
+# --- Discover invocation IDs from CLI / xcelerate logs --------------------
+
+rn_id=$(extract_log_id "React Native invocation ID" "$RN_CLI_LOG")
+ccache_id=$(extract_log_id "Ccache invocation ID" "$RN_CLI_LOG")
+
+XCELERATE_LOGS=$(find "${BITRISE_DEPLOY_DIR:-.}" -name 'xcelerate-*.log' 2>/dev/null || true)
+xcode_id=""
+if [ -n "$XCELERATE_LOGS" ]; then
+  # The xcode invocation ID is logged as "Registering invocation relation: parent=<rn>, child=<xcode>, build-tool=xcode".
+  xcode_id=$(grep -hoE "child=[a-zA-Z0-9-]+, build-tool=xcode" $XCELERATE_LOGS 2>/dev/null \
```
zsolt-marta-bitrise: [bitrise-plugin-ai-reviewer]: scripts/react_e2e_assert_invocations_be.sh:97:0:7d64ec9d

[//]: # ([skip ci])


**🐛 Bug**

The regex pattern and variable usage have two problems: (1) $XCELERATE_LOGS is unquoted, causing word splitting and glob expansion issues if filenames contain spaces or special characters. (2) The regex pattern 'child=[a-zA-Z0-9-]+, build-tool=xcode' doesn't match the actual log format. Based on xcodebuild.go, the format is 'parent=<id> → child=<id> (build-tool=xcode)' with an arrow and parentheses, not 'child=' and comma.

🔄 Suggestion:
```suggestion
  xcode_id=$(grep -hoE "child=[a-zA-Z0-9-]+ \(build-tool=xcode\)" "$XCELERATE_LOGS" 2>/dev/null \
    | head -1 \
    | sed -E 's/^child=([a-zA-Z0-9-]+).*/\1/')
```

zsolt-marta-bitrise (commented):

scripts/react_e2e_assert_invocations_be.sh
Badlazzor: Addressed in 8d77f14: dropped the xcelerate-log child-id extraction entirely (the regex was wrong AND `$XCELERATE_LOGS` was unquoted as you flagged). Children are now verified by querying the child-invocations endpoint directly and asserting the per-tool group is non-empty when that tool was active locally — no log parsing needed.

Also fixed the parent-invocation field names to match `BuildToolInvocationInfoPresenter#to_h` (`id`→`invocationId`, `buildTool`→`tool`, `workflow`→`workflowName`, `cacheHits/Misses`→`cacheHitRate`); those were the other failures from the first run.

scripts/react_e2e_assert_invocations_be.sh
```diff
@@ -0,0 +1,182 @@
+#!/usr/bin/env bash
```
zsolt-marta-bitrise: Can we put this into https://github.com/bitrise-io/build-cache-e2e-testing so that our continuous e2e testing also checks this (also add a workflow there)?

I also noticed this has a different format than existing assertions there, please consolidate that

Badlazzor: Moved + consolidated in https://github.com/bitrise-io/build-cache-e2e-testing/pull/3 — new `e2e-react-native-android` / `e2e-react-native-ios` workflows trigger `react-seek-*-e2e` remotely on the CLI app and assert via the existing `assert_field_value` / `assert_non_empty_field` / `fatal` helpers (no more emoji `ok/note/fail` counter). Children checked via a new `assert_invocation_has_child_react_native` that hits `/child-invocations.json`. Dropped the local script + workflow steps in 6350d01. Needs `RN_APP_START_TOKEN` set on the e2e-testing app before the new workflows can run.

PR #318, opened 2026-05-08
fix(ACI-4918): exclude no-activity children from RN combined hit rate

Badlazzor:
## Summary

[ACI-4918](https://bitrise.atlassian.net/browse/ACI-4918): The combined cache hit rate the `react-native run` wrapper reports on its parent invocation is severely understated when any child invocation does no cacheable work. Idle children contribute `0` to the hit-rate sum but still increment `ChildCount`, so e.g. one xcode child at 80% plus one idle ccache child reads out as 40%.

This PR drops idle children (`Total == 0`) from the aggregate and surfaces the count so the user can see what was excluded.

## Changes

- `pkg/common/childstats/childstats.go`
  - New `Summary.NoActivityCount` field.
  - `Aggregator.Compute()` skips entries with `Total == 0` before any other accounting (baseline, failed, per-tool counters, mean / weighted hit rate). Skipped entries land in `NoActivityCount` only — they don't pollute `BaselineCount` even when the entry is marked baseline.
- `pkg/reactnative/post_run_deps.go`
  - Logs the `NoActivityCount` next to the existing aggregate lines.
  - Adds a dedicated log line for the all-children-no-activity case so the absence of a "Cache hit rate" line has a reason on the build log.
  - Switched the aggregate logging chain from `if/else if` to `switch` (gocritic).
- `pkg/common/childstats/childstats_test.go`
  - Updated existing tests to write entries with `Total > 0` (the old fixtures were unrealistic — production writers always set `Total` when there's activity).
  - New tests:
    - `TestAggregator_Compute_ExcludesNoActivityChildren` — one active + one idle child; mean is 80% (not 40%) and per-tool breakdown excludes the idle tool.
    - `TestAggregator_Compute_AllNoActivity` — all idle → `ChildCount == 0`, `NoActivityCount == N`, empty `ByTool`.
    - `TestAggregator_Compute_BaselineWithNoActivityIsExcluded` — baseline + idle is excluded from both `ChildCount` AND `BaselineCount`.
    - Renamed `TestAggregator_Compute_LegacyEntriesDefaultToNotFailed` → `TestAggregator_Compute_LegacyEntriesAreSkippedAsNoActivity`. Legacy entries (pre-Total) unmarshal to `Total=0` and are now skipped — documents the new behavior.

## Notes / follow-ups

- The gradle plugin writer (and any other RN-side child writer) should populate `Total` for entries that DID do cache work. If a writer reports `HitRate > 0` but `Total == 0` due to a bug, this PR will silently drop the entry instead of pulling the mean toward it. That's safer (the value would have been meaningless without `Total` anyway), but worth a short audit on the plugin side. Flagged as a follow-up in the ticket.

## Test plan

- [ ] `go test -tags unit -race ./pkg/common/childstats/... ./pkg/reactnative/...` passes locally ✅.
- [ ] CI green on this PR.
- [ ] After merge: trigger `react-seek-android-e2e` / `react-seek-ios-e2e` manually and confirm:
  - The `Cache hit rate (avg of N child invocations)` line reports a non-zero rate when any active child has hits.
  - The new `N child invocation(s) excluded from the aggregate: no cache activity.` line appears when ccache had no activity (Android) or wasn't activated (iOS).

[ACI-4918]: https://bitrise.atlassian.net/browse/ACI-4918?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/common/childstats/childstats.go
+++ b/pkg/common/childstats/childstats.go
@@ -189,6 +189,12 @@ type Summary struct {
 	// SkippedCount is the number of entries excluded (malformed only).
 	SkippedCount int
 
+	// NoActivityCount is the number of entries excluded because the child
+	// did no cacheable work (Total == 0). They are excluded from
+	// MeanHitRate / WeightedHitRate / ChildCount / per-tool aggregates so
+	// that idle children don't pull the reported hit rate toward 0%.
+	NoActivityCount int
+
 	// FailedCount is the number of entries that reported Failed=true. Used
 	// by parent wrappers to fail the wrapper invocation when any child
 	// failed (e.g. an xcodebuild child errored even though the parent script
@@ -255,6 +261,21 @@ func (a *Aggregator) Compute() (Summary, error) {
 
 			if a.Logger != nil {
 				a.Logger.Warnf("Skipping malformed child stats entry %s: %v", entryPath, err)
+			}
+
+			continue
+		}
+
+		// Skip entries with no cacheable work — including them only drags
+		// MeanHitRate toward 0% without telling the user anything useful.
+		if entry.Total == 0 {
+			summary.NoActivityCount++
+
+			if a.Logger != nil {
+				a.Logger.Debugf(
+					"Excluding child invocation %s (build tool: %s) from aggregate: no cache activity",
+					entry.ChildInvocationID, entry.BuildTool,
+				)
 			}
 
 			continue
--- a/pkg/common/childstats/childstats_test.go
+++ b/pkg/common/childstats/childstats_test.go
@@ -94,15 +94,16 @@ func TestAggregator_Compute_SimpleMeanAndByTool(t *testing.T) {
 	setHome(t)
 
 	w := NewWriter()
-	require.NoError(t, w.Write(Entry{ParentInvocationID: "p", ChildInvocationID: "a", BuildTool: "gradle", HitRate: 0.2}))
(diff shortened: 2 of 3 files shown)


zsolt-marta-bitrise:
[//]: # ([skip ci])



## Summary
Excludes idle children (with no cache activity) from the React Native combined hit rate aggregate. Adds `NoActivityCount` field to track excluded children and logs the count separately so users can see what was filtered out.


## Walkthrough
| File | Summary |
|------|---------|
| pkg/common/childstats/childstats.go | Added `NoActivityCount` field to Summary struct. Modified `Compute()` to skip entries with `Total == 0` before any accounting (baseline, failed, per-tool counters, mean/weighted hit rate). Excluded entries only increment `NoActivityCount` and don't pollute `BaselineCount`. |
| pkg/reactnative/post_run_deps.go | Converted if/else if logging chain to switch statement. Added logging for `NoActivityCount` when active children exist. Added dedicated log line for all-children-no-activity case. |
| pkg/common/childstats/childstats_test.go | Updated existing tests to use realistic fixtures with `Total > 0`. Added new tests: `TestAggregator_Compute_ExcludesNoActivityChildren`, `TestAggregator_Compute_AllNoActivity`, `TestAggregator_Compute_BaselineWithNoActivityIsExcluded`, and renamed legacy test to document no-activity behavior. |

zsolt-marta-bitrise (approved):

PR #319, opened 2026-05-08
fix(ACI-4923): require RN activation+workspace; print slug-prefixed details URL

Badlazzor:
## Summary

Fixes [ACI-4923](https://bitrise.atlassian.net/browse/ACI-4923): the link the CLI prints after a React Native wrapper invocation save returns **404**.

```
[13:40:05] React Native invocation saved. Visit 👉 https://app.bitrise.io/build-cache/invocations/react-native/ffc0a374-…
```

## Root cause

Two layers:

1. **The URL itself is missing the workspace slug.** The public web route that resolves a workspace from the user session does not include `react-native` in its build-tool dispatch — gradle / xcode happen to redirect to a useful page from the slug-less URL but RN does not. The route that actually renders is the slug-prefixed form already used by the JSON BFF route:

   ```
   https://app.bitrise.io/<workspace_slug>/invocations/react-native/<id>
   ```

   Verified live (workspace `322a005426441b60`, an RN invocation from PR #317 e2e):

   | URL | Status |
   |---|---|
   | `…/build-cache/invocations/react-native/<id>` (old) | 302 → `/users/sign_in` (no resolution) |
   | `…/build-cache/<workspace_slug>/invocations/react-native/<id>` (this PR) | 200 |

2. **The runner pressed on even when the workspace slug was missing.** If `react-native run` is invoked without `activate react-native` (or with an activation that didn't carry a workspace ID through), there's no slug to build a correct URL — and no workspace context for analytics either. The previous behaviour silently emitted the broken slug-less URL.

## Changes

- `pkg/reactnative/runner.go`
  - New activation gate `isReactNativeReady()` at the start of `Run`: checks the RN marker (`~/.bitrise/cache/reactnative/config.json`) **and** that the multiplatform analytics config carries a non-empty `AuthConfig.WorkspaceID`.
  - If either is missing, emit `MsgRNNotActivated` (one warning) and execute the wrapped command unchanged — no IPC plumbing, no `BITRISE_INVOCATION_ID` injection, no post-run analytics. **The user's build never fails just because activation was skipped.**
- `pkg/reactnative/post_run_deps.go`
  - `MsgRNInvocationSaved` is now `…Visit 👉 %s` and takes a fully-formed URL.
  - `rnInvocationDetailsURL(workspaceSlug, invocationID)` builds the slug-prefixed URL. With the activation gate in place we no longer need a workspace-less fallback — callers only reach this once the slug is known.
- `pkg/reactnative/run_cmd_test.go`
  - `activateRNHome(t)` helper writes the RN marker + multiplatform config into a temp `HOME` so existing tests don't depend on the developer's local activation state. Wired into `TestRunner_Run` and `TestRunner_PostRunHook`.
  - New `TestRunner_BypassWhenNotActivated` covering both bypass paths: missing RN marker, and present marker with empty `WorkspaceID`. Asserts no invocation ID injected and `postRun` not fired.
- `pkg/reactnative/post_run_deps_test.go`
  - `TestRNInvocationDetailsURL_WithWorkspace` — slug-prefixed shape. (The previous `…FallsBackWhenWorkspaceMissing` case is gone since the fallback is gone.)

## Notes / out of scope

- The same slug-less URL is also emitted by `cmd/xcode/xcodebuild.go:46` (`MsgInvocationSaved`) and the bazel BES base in `internal/config/bazel/bazelrc.gotemplate`. Today gradle/xcode happen to redirect via session, so they're not user-visible failures — kept out of this PR per the scope of ACI-4923. Easy follow-up if/when the website tightens the redirect dispatch.

## Test plan

- [ ] `go test -tags unit -race ./pkg/reactnative/...` passes locally ✅.
- [ ] CI green on this PR.
- [ ] After merge: confirm a fresh `react-native run` build prints the slug-prefixed URL and that opening it in a browser renders the invocation page.
- [ ] Manual: invoke `react-native run` without first running `activate react-native` → expect the `MsgRNNotActivated` warning and the wrapped command running to completion unchanged.

[ACI-4923]: https://bitrise.atlassian.net/browse/ACI-4923?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/reactnative/post_run_deps.go
+++ b/pkg/reactnative/post_run_deps.go
@@ -24,11 +24,26 @@ import (
 // ---------------------------------------------------------------------------
 
 // MsgRNInvocationSaved is the log line emitted after the BE confirms the
-// React Native wrapper invocation was persisted. The URL points at the
-// per-invocation details page in the Bitrise app. Mirrors the equivalent
-// xcode and gradle messages so users (especially on non-Bitrise CI) can
-// jump straight to the invocation in the Bitrise UI.
-const MsgRNInvocationSaved = "React Native invocation saved. Visit 👉 https://app.bitrise.io/build-cache/invocations/react-native/%s"
+// React Native wrapper invocation was persisted. Format args: pre-formatted
+// URL pointing at the per-invocation details page in the Bitrise app.
+// Use rnInvocationDetailsURL to build the URL — the workspace slug is
+// required for the page to render (without it the route 404s for
+// react-native, see ACI-4923).
+const MsgRNInvocationSaved = "React Native invocation saved. Visit 👉 %s"
+
+// rnInvocationDetailsURL returns the public details URL for a React Native
+// invocation. When the workspace slug is known, the URL includes it
+// (`/build-cache/<workspace>/invocations/react-native/<id>`) and renders
+// the page directly. When it isn't, fall back to the workspace-less form;
+// it does not currently render for react-native, but it stays consistent
+// with the other tools' log lines and with future BE behaviour.
+func rnInvocationDetailsURL(workspaceSlug, invocationID string) string {
+	if workspaceSlug == "" {
+		return fmt.Sprintf("https://app.bitrise.io/build-cache/invocations/react-native/%s", invocationID)
+	}
+
+	return fmt.Sprintf("https://app.bitrise.io/build-cache/%s/invocations/react-native/%s", workspaceSlug, invocationID)
+}
 
 // postRunDeps handles post-run analytics: invocation reporting, ccache stats
 // collection, and invocation relation registration.
@@ -140,7 +155,7 @@ func (d *postRunDeps) run(ctx context.Context, wrapperInvocationID string, args 
 	} else {
 		// BE confirmed the invocation was stored — surface the details URL
 		// so users can jump to it from the build log.
-		d.logger.TInfof(MsgRNInvocationSaved, wrapperInvocationID)
+		d.logger.TInfof(MsgRNInvocationSaved, rnInvocationDetailsURL(d.authConfig.WorkspaceID, wrapperInvocationID))
 	}
 
 	if err := agg.Cleanup(); err != nil {
--- a/pkg/reactnative/post_run_deps.go
+++ b/pkg/reactnative/post_run_deps.go
@@ -32,16 +32,11 @@ import (
 const MsgRNInvocationSaved = "React Native invocation saved. Visit 👉 %s"
 
 // rnInvocationDetailsURL returns the public details URL for a React Native
-// invocation. When the workspace slug is known, the URL includes it
-// (`/build-cache/<workspace>/invocations/react-native/<id>`) and renders
-// the page directly. When it isn't, fall back to the workspace-less form;
-// it does not currently render for react-native, but it stays consistent
-// with the other tools' log lines and with future BE behaviour.
+// invocation. The workspace slug is required: the route 404s without it for
+// the react-native build-tool (gradle / xcode happen to redirect through
+// session, but RN does not). Callers must only invoke this once the runner
+// has confirmed activation is complete and the workspace slug is set.
 func rnInvocationDetailsURL(workspaceSlug, invocationID string) string {
-	if workspaceSlug == "" {
-		return fmt.Sprintf("https://app.bitrise.io/build-cache/invocations/react-native/%s", invocationID)
-	}
-
 	return fmt.Sprintf("https://app.bitrise.io/build-cache/%s/invocations/react-native/%s", workspaceSlug, invocationID)
 }
 
--- a/pkg/reactnative/post_run_deps_test.go
+++ b/pkg/reactnative/post_run_deps_test.go
@@ -1,0 +1,29 @@
+//go:build unit
+
+package reactnative
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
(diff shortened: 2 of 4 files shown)


zsolt-marta-bitrise:
[//]: # ([skip ci])



## Summary
Fixes ACI-4923: React Native invocation details URL now includes workspace slug to resolve correctly. Added readiness check to bypass RN wrapping when activation or workspace is missing, improving resilience.


## Walkthrough
| File | Summary |
|------|---------|
| pkg/reactnative/post_run_deps.go | Modified MsgRNInvocationSaved to use format string. Added rnInvocationDetailsURL() helper to build slug-prefixed URL when workspace available, falls back to slug-less form otherwise. Updated post-run logging to use the helper. |
| pkg/reactnative/runner.go | Added MsgRNNotActivated constant and isReactNativeReady() method to check activation completion and workspace ID presence. Modified Run() to bypass wrapping when not ready. Moved invocation ID generation after readiness check. |
| pkg/reactnative/post_run_deps_test.go, run_cmd_test.go | Added and updated tests for URL generation and readiness checks. |

pkg/reactnative/post_run_deps.go
```diff
@@ -24,11 +24,26 @@ import (
...
+// react-native, see ACI-4923).
+const MsgRNInvocationSaved = "React Native invocation saved. Visit 👉 %s"
+
+// rnInvocationDetailsURL returns the public details URL for a React Native
+// invocation. When the workspace slug is known, the URL includes it
+// (`/build-cache/<workspace>/invocations/react-native/<id>`) and renders
+// the page directly. When it isn't, fall back to the workspace-less form;
+// it does not currently render for react-native, but it stays consistent
+// with the other tools' log lines and with future BE behaviour.
+func rnInvocationDetailsURL(workspaceSlug, invocationID string) string {
```
zsolt-marta-bitrise, line 39: How can we not know the workspace slug? It's required during activation (it should get extracted from the JWT token too)

Badlazzor: Actually, if we want to be sure, we should add an activation check for the "run command". As it is, we might have a run command triggered without a workspace id on a users machine if they are running it before activate. 
This is not the place to actually test that, you are absolutely right. It was just the eager AI spotting that it might be empty.
Should I refactor that in this PR?

zsolt-marta-bitrise: Yeah an activation check makes sense. Can we do a noop fallback (i.e. not actually wrapping) with a warning log, so we don't cause failures?

Badlazzor: yeah thats what I was thinking too

Badlazzor: Done in 8cc7402.

`Runner.Run` now starts with an `isReactNativeReady()` gate that checks both the RN marker (`~/.bitrise/cache/reactnative/config.json`) and the multiplatform analytics config carrying a non-empty `AuthConfig.WorkspaceID`. If either is missing, we log a single warning (`MsgRNNotActivated` in `runner.go`) and execute the wrapped command unchanged — no IPC plumbing, no `BITRISE_INVOCATION_ID` injection, no post-run analytics. The user's build never fails just because activation was skipped.

With that gate in place, `rnInvocationDetailsURL` no longer needs the workspace-less fallback — it requires the slug and the only caller reaches it after the gate.

Tests added in `run_cmd_test.go` (`TestRunner_BypassWhenNotActivated`) cover both bypass paths: missing RN marker and present marker but empty `WorkspaceID`.

pkg/reactnative/post_run_deps.go
```diff
@@ -24,11 +24,21 @@ import (
...
+// required for the page to render (without it the route 404s for
+// react-native, see ACI-4923).
+const MsgRNInvocationSaved = "React Native invocation saved. Visit 👉 %s"
+
+// rnInvocationDetailsURL returns the public details URL for a React Native
+// invocation. The workspace slug is required: the route 404s without it for
+// the react-native build-tool (gradle / xcode happen to redirect through
+// session, but RN does not). Callers must only invoke this once the runner
+// has confirmed activation is complete and the workspace slug is set.
+func rnInvocationDetailsURL(workspaceSlug, invocationID string) string {
```
zsolt-marta-bitrise, line 39: [bitrise-plugin-ai-reviewer]: pkg/reactnative/post_run_deps.go:39:0:1f6f0081

[//]: # ([skip ci])


**🐛 Bug**

The function does not implement the fallback behavior described in the PR. According to the PR description, it should "fall back to the slug-less form otherwise so the log line stays useful even when the workspace is unknown." When workspaceSlug is empty, this will produce a malformed URL like "https://app.bitrise.io//invocations/react-native/...". The function should check if workspaceSlug is empty and return the slug-less URL format in that case.

🔄 Suggestion:
```suggestion
func rnInvocationDetailsURL(workspaceSlug, invocationID string) string {
	if workspaceSlug == "" {
		return fmt.Sprintf("https://app.bitrise.io/build-cache/invocations/react-native/%s", invocationID)
	}

	return fmt.Sprintf("https://app.bitrise.io/build-cache/%s/invocations/react-native/%s", workspaceSlug, invocationID)
}
```

zsolt-marta-bitrise (commented):

zsolt-marta-bitrise (approved):

PR #320, opened 2026-05-08
fix(ACI-4922): consolidate benchmark phase log into a single summary

Badlazzor:
## Summary

Fixes [ACI-4922](https://bitrise.atlassian.net/browse/ACI-4922): the activate step's log says \"Benchmark baseline mode: disabling cache and enabling analytics only\" but the build's gradle/xcode commands actually use the cache.

## Root cause (verified from the two repro builds)

The activator queries the BE for a benchmark phase **per build tool** (`gradle`, `xcode`). The BE rotates phases independently per `(workspace, app, workflow, build_tool)` and for the same RN workflow it returns different phases per tool:

| Workflow | gradle phase | xcode phase | Tool the workflow runs | \"baseline\" warning fired for |
|---|---|---|---|---|
| primary-android (build `5a3206fb…`) | warmup | baseline | gradle (caches normally) | xcode (irrelevant, no xcodebuild runs) |
| primary-ios (build `6c6a265a…`) | baseline | warmup | xcode (caches normally) | gradle (irrelevant, no gradle runs) |

Each per-tool `ApplyBenchmarkPhase` emitted its own warning, so the user saw \"baseline\" + \"cache being used\" and reported the contradiction. No actual cache-disable bug — the disable was applied correctly, just to the *unused* tool.

### Consolidated benchmark phase log

- `internal/config/common/benchmark_summary.go` (new): `LogBenchmarkSummary(logger, tools)` reads each tool's phase from its phase file and emits a single line:
  ```
  (i) Benchmark phases: gradle=warmup (cache enabled, hit rate may not be ideal), xcode=baseline (cache disabled)
  ```
  Uses `Warnf` (yellow) only when *every* active tool is in baseline. Otherwise stays at `Infof` so the line doesn't make the build look degraded when the tool the build actually runs is caching normally.
- `internal/config/gradle/benchmark.go`, `internal/config/xcelerate/benchmark.go`: drop the per-tool `(i) Benchmark phase: …`, `Benchmark baseline mode: …`, `(i) Benchmark warmup phase: …` lines. Param mutation, env export, file write behaviour unchanged.
- `cmd/gradle/activate_gradle.go`, `cmd/xcode/activate_xcode.go`, `pkg/reactnative/activator.go`: call `LogBenchmarkSummary` once at the end of activation with whichever tools were active.

### Tie ccache activation to gradle in the RN flow

ccache only meaningfully wraps the Android/Gradle native build path on RN, so its activation should follow gradle's. Two ties added in `pkg/reactnative/Activator`:

1. **Hard tie at construction** — `CppEnabled=true` + `GradleEnabled=false` skips ccache wiring entirely (no helper started). Logs the reason.
2. **Phase tie at activation** — when gradle ends up in benchmark baseline this run, skip the ccache activation + storage-helper start so the baseline measurement isn't polluted by a still-warm ccache. Stop-gap until [ACI-4926](https://bitrise.atlassian.net/browse/ACI-4926) lands proper per-tool benchmark phase support for ccache (BE's `command_benchmark_status` route's `:tool` regex only accepts `gradle|xcode|bazel` today).

### Tests

- `internal/config/common/benchmark_summary_test.go` covers the no-phase / mixed / all-baseline / single-tool / sort-stable cases.
- `pkg/reactnative/activator_test.go` covers the hard tie (cpp without gradle skipped at construction).

## Verification

- `go build ./... && go test -tags unit -race ./...` clean.
- Logging-only fix for the user-perceived contradiction; param mutation behaviour and the values written to env vars / phase files are unchanged.

## Out of scope (separate follow-ups)

- **Latent gradle.properties bug**: `internal/config/gradle/activate.go:40` calls `UpdateGradleProps(params, …)` with the un-mutated outer `params`, so when `ApplyBenchmarkPhase` flips `Cache.Enabled = false` the init script picks up the disable but `gradle.properties` still gets `org.gradle.caching=true`. Doesn't manifest in either of the two repro builds (the workflow's relevant tool is in warmup in both cases) but would bite a gradle-running workflow that actually lands in baseline. Worth a separate small PR.
- **BE rotation divergence**: per-tool independent rotation for the same workflow is user-confusing even with the new log; worth a conversation with whoever owns the website rotation logic.
- **ccache benchmark phase support** — tracked in [ACI-4926](https://bitrise.atlassian.net/browse/ACI-4926).

[ACI-4922]: https://bitrise.atlassian.net/browse/ACI-4922?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ACI-4926]: https://bitrise.atlassian.net/browse/ACI-4926?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cmd/gradle/activate_gradle.go
+++ b/cmd/gradle/activate_gradle.go
@@ -63,6 +63,8 @@ If the "# [start/end] generated-by-bitrise-build-cache" block is already present
 			return fmt.Errorf("activate plugins for Gradle: %w", err)
 		}
 
+		configcommon.LogBenchmarkSummary(logger, []string{configcommon.BuildToolGradle})
+
 		logger.TInfof("✅ Bitrise plugins activated")
 
 		return nil
--- a/cmd/xcode/activate_xcode.go
+++ b/cmd/xcode/activate_xcode.go
@@ -7,6 +7,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/bitrise-io/bitrise-build-cache-cli/v2/cmd/common"
+	configcommon "github.com/bitrise-io/bitrise-build-cache-cli/v2/internal/config/common"
 	"github.com/bitrise-io/bitrise-build-cache-cli/v2/internal/config/xcelerate"
 	"github.com/bitrise-io/bitrise-build-cache-cli/v2/internal/utils"
 )
@@ -47,7 +48,7 @@ This command will:
 		activateXcodeParams.DebugLogging = common.IsDebugLogMode
 		logger.Infof("Activate Xcode params: %+v", activateXcodeParams)
 
-		return xcelerate.Activate(
+		if err := xcelerate.Activate(
 			cmd.Context(),
 			logger,
 			utils.DefaultOsProxy{},
@@ -56,7 +57,13 @@ This command will:
 			utils.DefaultDecoderFactory{},
 			activateXcodeParams,
 			utils.AllEnvs(),
-		)
+		); err != nil {
+			return err //nolint:wrapcheck // xcelerate.Activate already returns wrapped errors
+		}
+
+		configcommon.LogBenchmarkSummary(logger, []string{configcommon.BuildToolXcode})
+
+		return nil
 	},
 }
 
--- a/internal/config/common/benchmark_summary.go
+++ b/internal/config/common/benchmark_summary.go
@@ -28,10 +28,8 @@ func LogBenchmarkSummary(logger log.Logger, tools []string) {
 		phase string
 	}
 
-	var (
-		entries         []entry
-		anyCacheEnabled bool
-	)
+	entries := make([]entry, 0, len(tools))
+	anyCacheEnabled := false
 
 	for _, tool := range tools {
 		phase := ReadBenchmarkPhaseFile(tool, logger)
--- a/internal/config/common/benchmark_summary.go
+++ b/internal/config/common/benchmark_summary.go
@@ -1,0 +1,80 @@
+package common
+
+import (
+	"sort"
+	"strings"
+
+	"github.com/bitrise-io/go-utils/v2/log"
+)
+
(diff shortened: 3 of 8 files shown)


zsolt-marta-bitrise:
[//]: # ([skip ci])



## Summary
Consolidates benchmark phase logging into a single summary line across all build tools (Gradle, Xcode, ccache) to eliminate confusing per-tool warnings. Ties ccache activation to Gradle in React Native flows and skips ccache when Gradle is in baseline phase to avoid polluting measurements.


## Walkthrough
| File | Summary |
|------|---------|
| internal/config/common/benchmark_summary.go | New file: LogBenchmarkSummary() reads each tool's phase file and emits single consolidated log line showing all active tools' phases and cache status. Logs at warn level only when all tools are in baseline. |
| internal/config/gradle/benchmark.go, internal/config/xcelerate/benchmark.go | Removed per-tool log lines from ApplyBenchmarkPhase(). Functions now only handle parameter mutation and env var/file exports; logging delegated to LogBenchmarkSummary(). |
| cmd/gradle/activate_gradle.go, cmd/xcode/activate_xcode.go | Added LogBenchmarkSummary() call after successful activation to emit consolidated benchmark phase summary. |
| pkg/reactnative/activator.go | Hard tie: skip ccache wiring if Gradle disabled. Phase tie: skip ccache activation when Gradle in baseline. Added LogBenchmarkSummary() call at end of activation. |
| internal/config/common/benchmark_summary_test.go, pkg/reactnative/activator_test.go | Added/updated tests covering no-phase, mixed phases, all-baseline, single-tool cases and hard tie (cpp without gradle). |

zsolt-marta-bitrise (approved):

PR #323, opened 2026-05-13
fix(release): make Homebrew tap publish non-fatal

Badlazzor:
## Summary

Last CLI release tripped on goreleaser's homebrew step (`404 PUT /Formula/bitrise-build-cache.rb` against `bitrise-io/homebrew-bitrise-build-cache`) and that fatal exit blocked the rest of the workflow — including the GAR mirror upload and downstream steplib auto-updates. The GH release binaries (the critical path used by every Bitrise build via `installer.sh`) had already landed before the failure, so the user-facing impact was limited, but the workflow status was red and the steplib promotion didn't run.

## Critical-path ordering

| Channel | Critical? |
|---|---|
| GH release tarballs + checksums (`installer.sh` consumes these) | **Yes** — every Bitrise build |
| GAR mirror of the same tarballs (installer fallback) | **Yes** — every Bitrise build when GH is degraded |
| Homebrew tap (`bitrise-io/homebrew-bitrise-build-cache`) | No — nice-to-have, not in the install path |

## Changes

- `bitrise.yml` — split the release script:
  - **"Goreleaser (create binaries + publish to GH)"** now runs `goreleaser release --skip=homebrew`. Binaries + GH release publish without touching the tap.
  - **GAR mirror steps** unchanged (now guaranteed to run even when the tap is broken).
  - **New "Publish Homebrew formula (non-fatal)"** step re-runs goreleaser with every other phase skipped so only the brew publisher executes, reusing the `dist/` artefacts from the first run. `is_skippable: true` ensures a tap-permission failure here can't fail the workflow.
- `.claude/skills/release/SKILL.md` — document the rule under the existing "Critical path" section so future releases don't treat a brew failure as a drop-everything incident.

## Follow-up (out of scope for this PR)

Root-cause the 404: the bot user behind `GIT_BOT_USER_ACCESS_TOKEN` is not a collaborator with push access on `bitrise-io/homebrew-bitrise-build-cache`. Once added, the brew step should succeed again — but the defensive split above means failures from here on are recoverable instead of release-blocking.

## Test plan

- [ ] Next CLI tag triggers the release workflow; verify the "Goreleaser (create binaries + publish to GH)" step succeeds without touching the tap.
- [ ] Verify GAR upload completes.
- [ ] Verify "Publish Homebrew formula (non-fatal)" runs and either succeeds (once the bot has tap access) or fails without failing the workflow.

### DIFF
--- a/.claude/skills/release/SKILL.md
+++ b/.claude/skills/release/SKILL.md
@@ -23,6 +23,17 @@ Concrete rules:
 - **Treat any failure of the CLI `release` workflow as a drop-everything-and-fix incident.** Don't move on to step releases until the CLI release workflow is green AND the v2.6.x release has all 8 expected assets (6 platform tarballs, checksums.txt, both verification XMLs).
 - **Smoke-test installer.sh edits on a real Bitrise build before merging.** The release flow does not regenerate this file, and there is no automated test that exercises it under `/bin/sh`.
 
+### Brew tap is best-effort, NOT critical path
+
+`bitrise-io/homebrew-bitrise-build-cache` is a nice-to-have publication target, not part of the install flow used by Bitrise builds. The "Publish Homebrew formula" step in `bitrise.yml`'s `release` workflow runs goreleaser with only the brew publisher and is marked `is_skippable: true`, so a tap-permission failure (e.g. 404 from `PUT /Formula/bitrise-build-cache.rb`) won't break the release.
+
+If the brew step fails:
+
+- Confirm the release workflow's overall status is still green.
+- Verify the GitHub release has all 8 expected assets and the GAR mirror upload succeeded — these are critical.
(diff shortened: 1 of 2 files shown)


zsolt-marta-bitrise:
[//]: # ([skip ci])



## Summary
Split the release workflow to make Homebrew tap publishing non-fatal. Goreleaser now skips the brew step on the first run, then a separate skippable step re-runs it with only the brew publisher enabled. This ensures GitHub release binaries and GAR mirror uploads complete even if the tap publish fails.


## Walkthrough
| File | Summary |
|------|---------|
| bitrise.yml | Modified Goreleaser step to skip homebrew with --skip=homebrew flag. Added new "Publish Homebrew formula (non-fatal)" step that re-runs goreleaser with only brew publisher enabled, marked is_skippable: true |
| .claude/skills/release/SKILL.md | Documented that Homebrew tap is best-effort and not part of the critical install path, explaining how to handle brew step failures |

zsolt-marta-bitrise (approved):

PR #331, opened 2026-05-15
ACI-4939: runtime opt-out for Gradle mirrors init script

zsolt-marta-bitrise:
## Summary

Follow-up from the [Maven Central rate-limit postmortem](https://bitrise.atlassian.net/wiki/spaces/INCIDENT/pages/4980998155) ([ACI-4939](https://bitrise.atlassian.net/browse/ACI-4939), parent epic ACI-4935).

Today's customer opt-out is a shared script step that customers paste at the start of their workflow: it `export`s `BITRISE_MAVENCENTRAL_PROXY_ENABLED=false`, calls `envman add` for the same key, and `rm`s `~/.gradle/init.d/bitrise-gradle-mirrors.init.gradle.kts`. Raw shell pasted into a workflow is fragile, doesn't survive workflow rewrites, and forces every opt-out request through CSM 1-on-1.

This change makes the installed init script re-read `BITRISE_MAVENCENTRAL_PROXY_ENABLED` inside the Gradle JVM. When the env var is explicitly `"false"`, the script logs and returns early — no replacements are wired up. Customers can now disable the mirrors per workflow / per workspace by setting the env var via:

- Secrets (workspace or app-level)
- Step inputs (`envs:` on the relevant step)
- App-level env vars
- An inline script step (`envman add --key BITRISE_MAVENCENTRAL_PROXY_ENABLED --value false`)

No more `rm` of the init file. The activation-time semantics are unchanged — the file is still only written when `BITRISE_MAVENCENTRAL_PROXY_ENABLED == "true"` at activate time — so the runtime guard is a *override layer* on top, satisfying the postmortem agreement that "workflow-level overrides take precedence over the monolith's LD-based default".

### Runtime semantics

Only the literal `"false"` disables. Unset, `"true"`, or anything else leaves the mirrors active. Rationale: the file's existence already encodes "the operator decided to enable" at activate time; the runtime check is just an explicit opt-out hatch.

### Changes

- `internal/config/gradle/mirrors/asset/gradle-mirrors.init.gradle.kts.gotemplate` — early-return guard at the top of `InternalRepositoryPlugin.apply()`.
- `internal/config/gradle/mirrors/activate_test.go` — asserts the guard renders into the AMS1-all-mirrors fixture.
- `cmd/gradle/activate_gradle_mirrors.go` — cobra `Long` help mentions the runtime opt-out.
- `pkg/gradle/mirrors/activator.go` — `Activate` godoc mentions the runtime opt-out.

### Out of scope

- Devcenter page documenting the opt-out (item 2 of the postmortem plan) — separate work, outside this repo.

## Test plan

- [x] `make test-unit` — passes locally
- [x] `make lint` — clean
- [ ] CI: verify `cache-gradle-test` workflow still picks up the MavenCentral and Google mirrors with `BITRISE_MAVENCENTRAL_PROXY_ENABLED=true` (the existing e2e checks at `bitrise.yml:1046` cover this — the guard's early-return only fires for `"false"`).
- [ ] Manual: in a sandbox app, install the init script, set `BITRISE_MAVENCENTRAL_PROXY_ENABLED=false` via Secrets, run a Gradle build, confirm the `[Bitrise Gradle Mirrors] ... skipping ... activation` log line appears and that no mirror URLs are substituted.

[ACI-4939]: https://bitrise.atlassian.net/browse/ACI-4939?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cmd/gradle/activate_gradle_mirrors.go
+++ b/cmd/gradle/activate_gradle_mirrors.go
@@ -20,7 +20,11 @@ When no flags are provided, all known mirrors are enabled.
 
 The command checks the BITRISE_MAVENCENTRAL_PROXY_ENABLED environment variable
 and only installs the init script when it is set to "true".
-The mirror URL is determined by the BITRISE_DEN_VM_DATACENTER environment variable.`,
+The mirror URL is determined by the BITRISE_DEN_VM_DATACENTER environment variable.
+
+The installed init script re-reads BITRISE_MAVENCENTRAL_PROXY_ENABLED at Gradle
+build time: setting it to "false" (via Secrets, step inputs, or app config)
+disables the mirrors per workflow / per workspace without removing the file.`,
 	SilenceUsage: true,
 	RunE: func(cmd *cobra.Command, _ []string) error {
 		activator := mirrorspkg.NewActivator(mirrorspkg.ActivatorParams{
--- a/internal/config/gradle/mirrors/activate_test.go
+++ b/internal/config/gradle/mirrors/activate_test.go
@@ -68,6 +68,8 @@ func TestActivate(t *testing.T) {
 				"https://repository-manager-ams.services.bitrise.io:8090/maven/gradle-plugins",
 				`r.getUrl().toString().trimEnd('/').equals("https://plugins.gradle.org/m2")`,
 				`"[Bitrise Gradle Mirrors] [$ts] $message"`,
+				`if (System.getenv("BITRISE_MAVENCENTRAL_PROXY_ENABLED") == "false") {`,
+				`log("BITRISE_MAVENCENTRAL_PROXY_ENABLED=false, skipping Bitrise Gradle mirror activation")`,
 				`log("beforeSettings fired,`,
 				`log("settingsEvaluated fired,`,
 				`if (getParent() == null) {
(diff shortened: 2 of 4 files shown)


zsolt-marta-bitrise:
[//]: # ([skip ci])



## Summary
Implements runtime opt-out for Gradle mirrors by adding an early-return guard in the init script that checks `BITRISE_MAVENCENTRAL_PROXY_ENABLED` at build time. Customers can now disable mirrors per workflow via Secrets, step inputs, or app config without removing the init file.


## Walkthrough
| File | Summary |
|------|---------|
| gradle-mirrors.init.gradle.kts.gotemplate | Added early-return guard in InternalRepositoryPlugin.apply() that skips mirror configuration when BITRISE_MAVENCENTRAL_PROXY_ENABLED is explicitly "false" |
| activate_gradle_mirrors.go, activator.go | Updated help text and godoc to document runtime opt-out behavior and env var configuration options |
| activate_test.go | Added test assertions verifying the runtime opt-out guard is present in generated init script |

Badlazzor (approved):

PR #333, opened 2026-05-15
ACI-4952: auto-inject EAS_LOCAL_BUILD_WORKINGDIR for Expo eas build

Badlazzor:
## Summary

Expo's `eas build --local` picks a fresh tmp working directory on every invocation, which busts every cache that hashes absolute source paths (Gradle, Xcode, ccache). The documented workaround is to pin `EAS_LOCAL_BUILD_WORKINGDIR` to a stable path; this PR applies it automatically so customers don't have to.

- **CI path** — `Activator.Activate()` now exports `EAS_LOCAL_BUILD_WORKINGDIR` via the existing `envexport` pipeline (envman + GITHUB_ENV + current process) on any detected CI (Bitrise / GitHub Actions / CircleCI / GitLab). User-supplied values always win.
- **Local path** — `Runner.Run()` detects `eas build` (direct, or via `npx` / `pnpm` / `yarn` / `bunx` / `bun`) and injects the env var into the child process only — no host-state side-effects on developer machines.
- **Path picker** — `/Users/vagrant/build` on Bitrise (matches ticket / example workflow), `$HOME/build` elsewhere.
- `DetectCIProvider` exported from `internal/config/common` so the activator and Runner can share the existing detection logic.

## Test plan
- [x] `make check` clean (lint 0 issues, full unit test suite green)
- [x] New unit tests in `pkg/reactnative/eas_test.go`, `run_cmd_test.go`, `activator_test.go` cover:
  - direct + package-manager-wrapped `eas build` detection (positive + negative cases)
  - Bitrise vs other CI vs local path selection
  - user-supplied `EAS_LOCAL_BUILD_WORKINGDIR` preservation
  - CI vs non-CI activation gating
- [ ] e2e: trigger an Expo RN pipeline on a Bitrise stack and confirm the env var lands in the build env + cache hit rate improves across consecutive runs.

Ticket: https://bitrise.atlassian.net/browse/ACI-4952

### DIFF
--- a/pkg/reactnative/eas.go
+++ b/pkg/reactnative/eas.go
@@ -2,8 +2,6 @@ package reactnative
 
 import (
 	"strings"
-
-	configcommon "github.com/bitrise-io/bitrise-build-cache-cli/v2/internal/config/common"
 )
 
 // EASWorkingDirEnv is the env var EAS Build reads to decide where to stage the
@@ -11,11 +9,6 @@ import (
 // otherwise, which defeats any source-path-sensitive cache (Gradle, Xcode,
 // ccache all hash absolute paths).
 const EASWorkingDirEnv = "EAS_LOCAL_BUILD_WORKINGDIR"
-
-// bitriseDefaultEASWorkingDir matches the convention documented in
-// ACI-4952 / the Bitrise example workflow. The vagrant user's HOME is
-// /Users/vagrant on Bitrise macOS stacks.
-const bitriseDefaultEASWorkingDir = "/Users/vagrant/build"
 
 // IsEASBuildInvocation reports whether the wrapped command is `eas build ...`,
 // either invoked directly or through a common JS package-manager runner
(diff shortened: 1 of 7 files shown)


pkg/reactnative/eas.go
```diff
@@ -0,0 +1,81 @@
...
+// EASWorkingDirEnv is the env var EAS Build reads to decide where to stage the
+// local build working directory. EAS picks a random tmp dir per invocation
+// otherwise, which defeats any source-path-sensitive cache (Gradle, Xcode,
+// ccache all hash absolute paths).
+const EASWorkingDirEnv = "EAS_LOCAL_BUILD_WORKINGDIR"
+
+// bitriseDefaultEASWorkingDir matches the convention documented in
+// ACI-4952 / the Bitrise example workflow. The vagrant user's HOME is
+// /Users/vagrant on Bitrise macOS stacks.
+const bitriseDefaultEASWorkingDir = "/Users/vagrant/build"
```
zsolt-marta-bitrise: This can change, why not `~/...` ?

pkg/reactnative/runner.go
```diff
@@ -225,6 +228,29 @@ func (r *Runner) ensureHelper(ctx context.Context, wrapperInvocationID string) {
...
 
+// maybeInjectEASWorkingDir pins EAS Build's local working directory to a
+// stable path when the wrapped command is `eas build` (directly or via a
+// package-manager runner). Without this, EAS picks a new tmp dir per
+// invocation and every downstream cache (Gradle, Xcode, ccache) misses
+// because absolute source paths feed into the cache keys.
+//
+// The injection is a no-op when the user has already set
+// EAS_LOCAL_BUILD_WORKINGDIR — explicit user intent wins.
+func (r *Runner) maybeInjectEASWorkingDir(environ []string, name string, cmdArgs []string) []string {
```
zsolt-marta-bitrise: Why do we use slices instead of map? I'd sugget using map everywhere

pkg/reactnative/eas.go
Badlazzor: Good call — switched to `$HOME/build` everywhere in a1b146e. `DefaultEASWorkingDir` now derives from HOME uniformly (returns "" when HOME is missing so callers skip rather than fall back to a hardcoded path). Today on Bitrise stacks that still resolves to `/Users/vagrant/build`, but the convention is no longer baked into the code.

pkg/reactnative/runner.go
Badlazzor: Refactored in a1b146e — `Run()` now converts `environ` to a map once on entry to the wrapped path, applies BITRISE_INVOCATION_ID + EAS workdir mutations on the map, and materialises back to a slice only at the `execFn` boundary (where the os/exec API requires `[]string`). `maybeInjectEASWorkingDir` now operates on the map directly, `environContains` is gone, and a `mapToEnviron` helper + roundtrip test was added. The bypass branch (when RN isn't activated) keeps its slice shape so the existing "environ unchanged when bypassed" contract still holds.

zsolt-marta-bitrise:
[//]: # ([skip ci])



## Summary
Auto-inject EAS_LOCAL_BUILD_WORKINGDIR for Expo eas build to stabilize cache paths. On CI, export via envman/GITHUB_ENV. On local machines, inject into child process only. Reuses existing CI detection logic.


## Walkthrough
| File | Summary |
|------|---------|
| internal/config/common/cache_config.go | Exported DetectCIProvider() function (renamed from detectCIProvider) for reuse by activator and runner |
| pkg/reactnative/eas.go | New file with EAS Build detection (IsEASBuildInvocation), working directory selection (DefaultEASWorkingDir), and environment conversion utilities (environToMap, mapToEnviron) |
| pkg/reactnative/activator.go | Added exportEASWorkingDirIfCI() to export EAS_LOCAL_BUILD_WORKINGDIR on detected CI platforms |
| pkg/reactnative/runner.go | Added maybeInjectEASWorkingDir() to inject EAS_LOCAL_BUILD_WORKINGDIR for eas build commands on local machines; switched to map-based environment handling |
| pkg/reactnative/{activator,eas,run_cmd}_test.go | New and updated unit tests covering EAS detection, CI vs local path selection, and user-supplied env var preservation |

pkg/reactnative/runner.go
zsolt-marta-bitrise: I actually meant the full react native flow, but if that's too much of a scope expansion, I suggest to follow up with that later

pkg/reactnative/eas.go
zsolt-marta-bitrise: Btw we have UserHomeDir in OsProxy, can't we use that?

zsolt-marta-bitrise (approved):
Still some outstanding items I'd like to address, but we can unblock the customer and followup later

PR #335, opened 2026-05-15
fix: ACI-4793 enrich downloaded file hash mismatch error with diagnostic context

duplabe:
## Summary

When `DownloadStream` (`internal/build_cache/kv/download.go`) detects a SHA256 divergence between the bytes it streamed and the server-supplied `x-flare-blob-validation-sha256`, the error currently reads:

```
downloaded file hash mismatch: expected <hex>, got <hex>
```

That's enough to know a hash diverged, but nothing else — not the key, not how much was received, not whether the wire was re-stitched across retries, and no IDs to grep on the server side.

This change enriches the error with cheap, already-in-scope context:

```
downloaded file hash mismatch for key "<key>": expected <hex>, got <hex> (received N bytes across M attempt(s), retried offset O, cache-operation-id="<id>", invocation-id="<id>")
```

Fields added:
- **key** — most important, especially for `DownloadStream` callers (xcelerate proxy, metadata JSON fetches) that don't otherwise log the key on failure
- **received N bytes** — distinguishes "got 0", "got partial", "got full size but corrupt"
- **M attempt(s)** + **retried offset** — flags retry-boundary stitching as a possible corruption source (the hasher continues across retries with `offset += n` on partial reads)
- **cache-operation-id** / **invocation-id** — already on `Client`, sent up as request headers; echoing them in the error lets ops correlate across client and server logs

### Explicitly not added
- Server response metadata (`kvReader.Metadata()` dump) — gRPC response headers can carry tokens or other sensitive values; an allowlist by prefix is brittle.
- `x-flare-blob-validation-level` — we send this as a *request* header on PUT, but the GET path doesn't send it and we have no evidence the server echoes it on GET responses.

## Test plan

- [x] `go test -tags unit -race ./internal/build_cache/kv/...` — passes
- [x] `make lint` — 0 issues
- [x] `TestClient_DownloadStream_MismatchValidation` extended to assert the new substrings (key, expected hash, byte count, attempt count) appear in the error

The error remains a plain `fmt.Errorf` (no sentinel wrapping), so existing `errors.Is(err, ErrCacheNotFound)` / `errors.Is(err, ErrCacheUnauthenticated)` checks at call sites are unaffected.

### DIFF
--- a/internal/build_cache/kv/download.go
+++ b/internal/build_cache/kv/download.go
@@ -88,12 +88,15 @@ func (c *Client) DownloadFile(ctx context.Context, filePath, key string, fileMod
 
 func (c *Client) DownloadStream(ctx context.Context, destination io.Writer, key string) error {
 	var offset int64
+	var totalBytes int64
+	var attempts uint
 
 	hasher := sha256.New()
 	multiWriter := io.MultiWriter(hasher, destination)
 	expectedHash := ""
 
 	downloadErr := retry.Times(c.downloadRetry).Wait(c.downloadRetryWait).TryWithAbort(func(attempt uint) (error, bool) {
+		attempts = attempt + 1
 		if attempt == 0 {
 			c.logger.TDebugf("Downloading %s", key)
 		} else {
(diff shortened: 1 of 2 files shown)


zsolt-marta-bitrise:
[//]: # ([skip ci])



## Summary
Enriches the hash mismatch error message in DownloadStream with diagnostic context including the cache key, byte count, attempt count, retry offset, cache operation ID, and invocation ID to aid troubleshooting and server-side correlation.


## Walkthrough
| File | Summary |
|------|---------|
| internal/build_cache/kv/download.go | Added tracking of total bytes received and attempt count; enriched hash mismatch error with key, byte count, attempt count, retry offset, cache-operation-id, and invocation-id |
| internal/build_cache/kv/download_test.go | Updated test expectations to assert new error message substrings |

ramil-bitrise (approved):

PR #344, opened 2026-05-27
ci: ACI-5007 automate R2 mirror + preboot startup-script bump on release

Badlazzor:
## Summary

Closes [ACI-5007](https://bitrise.atlassian.net/browse/ACI-5007). Removes the two manual chores after every CLI release:

- **R2 upload** — mirrors the four platform tarballs + checksums to the Cloudflare R2 bucket `build-cache-cli-releases` (origin for the preboot host-VM cache proxy at `http://${SUBNET_IP}1:59020/build-cache-cli-releases/<tarball>`). Flat layout (object key == filename), describe-or-upload immutability via `aws s3api head-object`.
- **Preboot bump PR** — sed-bumps `BITRISE_BUILD_CACHE_CLI_VERSION` + per-arch sha256 in both `startup_script_extension_{linux,macos}_bitvirt.sh` files in `bitrise-io/build-prebooting-deployments`, opens a PR as Bitrise Infrabot, watches the deployments-repo CI, then explicitly merges (bot is a bypass actor on `production` — GH `--auto` is NOT used because it doesn't honour bypass). After sed, asserts the working tree shows exactly `+2/-2` per file as a defensive check against a loose regex.

Pipeline now: `release → verify-release → bump-prebooting`. Each workflow is an independent build so they can be re-run individually without re-cutting the tag.

## Manual prerequisites before this works on a release

The CI changes alone are not enough — these provisioning steps must be done out-of-band before the next tag triggers the pipeline. Without them the new steps will fail on first run.

### 1. Cloudflare R2 API token

Create a token in the Cloudflare dashboard (R2 → Manage R2 API Tokens → Create API Token):
- **Permission**: `Object Read & Write`
- **Bucket scope**: `build-cache-cli-releases` only (do NOT use a broad/all-buckets token)
- **TTL**: no expiration (rotated manually)
- Name suggestion: `build-cache-cli-release-uploader`

Add as Bitrise app secrets on the `bitrise-build-cache-cli` app:
- `R2_ACCOUNT_ID=a484c7653eeba8c8c00a4bf3967860a3`
- `R2_ACCESS_KEY_ID`
- `R2_SECRET_ACCESS_KEY`

### 2. Bitrise Infrabot GitHub PAT for prebooting repo

Create a fine-grained PAT for the `Bitrise Infrabot` GitHub account scoped to `bitrise-io/build-prebooting-deployments`:
- `Contents: Read and write`
- `Pull requests: Read and write`
- `Metadata: Read-only` (implicit)
- TTL: no expiration (rotated manually)

Add as Bitrise app secret:
- `PREBOOTING_BOT_TOKEN`

### 3. Branch-protection bypass on prebooting repo

On `bitrise-io/build-prebooting-deployments` → Settings → Rules → ruleset for `production`:
- Add `Bitrise Infrabot` as a **bypass actor**.
- The script does NOT rely on `--auto` merge mode (which ignores bypass) — it watches CI then merges explicitly from the bot, so bypass-actor permission is what lets the explicit merge clear required-review gates.

## Notes

- Both R2 and GAR mirrors continue to run in parallel — GAR stays as the public versioned fallback for the installer.sh path and the startup-script's secondary download URL; R2 is the new primary origin for the host-VM cache proxy.
- Only `linux_amd64` + `darwin_arm64` are bumped in the preboot startup scripts — those are the only preboot VM architectures today.
- R2's S3 API rejects AWS CLI v2's default flexible checksums, so the upload script forces `AWS_REQUEST_CHECKSUM_CALCULATION=when_required` / `AWS_RESPONSE_CHECKSUM_VALIDATION=when_required`.
- Slack alerts (`#team-advanced-ci-alerts-website`) fire on either new step failing; the release itself is never re-run on a bump failure — retry the failed workflow standalone.

## Test plan

- [ ] All four secrets provisioned on the Bitrise app
- [ ] Bitrise Infrabot added as bypass actor on prebooting `production` rule
- [ ] Cut a no-op patch release; observe:
  - [ ] R2 bucket gets the four tarballs + checksums file under their bare filenames
  - [ ] PR opens against `bitrise-io/build-prebooting-deployments` with the correct version + per-arch sha256 bumps
  - [ ] Script watches deployments-repo CI, then merges the PR (squash, delete branch)
- [ ] Re-run the `release` workflow on the same tag; confirm R2 head-object skip-if-present prevents duplicate uploads
- [ ] Re-run `bump-prebooting` standalone on the same tag; confirm the script no-ops gracefully (no-changes guard)

[ACI-5007]: https://bitrise.atlassian.net/browse/ACI-5007?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


scripts/prebooting_pr_bump.sh
```diff
@@ -0,0 +1,128 @@
...
+# 3. Bump the three constants (CLI_VERSION on both, plus the per-arch SHA).
+#    Uses GNU sed (linux-docker stack). The bash literal `local` prefix
+#    keeps the regex narrow so we never match unrelated assignments.
+sed -i -E "s|^([[:space:]]*local BITRISE_BUILD_CACHE_CLI_VERSION=\")[^\"]+(\")|\1${tag}\2|" "$linux_script"
+sed -i -E "s|^([[:space:]]*local BITRISE_BUILD_CACHE_CLI_LINUX_AMD64_SHASUM=\")[^\"]+(\")|\1${linux_sha}\2|" "$linux_script"
+
+sed -i -E "s|^([[:space:]]*local BITRISE_BUILD_CACHE_CLI_VERSION=\")[^\"]+(\")|\1${tag}\2|" "$macos_script"
+sed -i -E "s|^([[:space:]]*local BITRISE_BUILD_CACHE_CLI_DARWIN_ARM64_SHASUM=\")[^\"]+(\")|\1${darwin_sha}\2|" "$macos_script"
+
+# Sanity-check the bumps actually landed.
```
zsolt-marta-bitrise, line 85: Can we also check that nothing else changed?

scripts/prebooting_pr_bump.sh
```diff
@@ -0,0 +1,128 @@
...
+)
+
+gh pr create \
+  --repo "$REPO" \
+  --base "$BASE" \
+  --head "$BRANCH" \
+  --title "chore: bump bitrise-build-cache CLI to ${tag_v}" \
+  --body "$pr_body"
+
+gh pr merge "$BRANCH" --repo "$REPO" --auto --squash
```
zsolt-marta-bitrise: Will the auto merge work with bypass? AFAIK it only works with explicit `git merge --bypass`

scripts/prebooting_pr_bump.sh
Badlazzor: Good catch — added in 7cb84f2. After the sed bumps the script now asserts `git diff --numstat` shows exactly `+2/-2` for each startup script and no other files modified. Any unintended match anywhere (or no match) fails the run before push.

scripts/prebooting_pr_bump.sh
Badlazzor: Confirmed your concern — `--auto` runs as a GitHub background process that does NOT honour bypass actors, so it would block on any required reviewer. Switched in 7cb84f2 to `gh pr checks --watch --fail-fast` followed by an explicit `gh pr merge --squash --delete-branch` from the Bitrise Infrabot. Because the bot is a bypass actor on `production`, the explicit merge clears required-review rules at merge time.

Side-effect: dropped the "Allow auto-merge" item from the prereq list — no longer needed.

zsolt-marta-bitrise (approved):

PR #350, opened 2026-06-03
feat: ACI-5028 store auth credentials in OS keychain

Badlazzor:
## Summary

- New package `internal/auth/keychain` — Get/Set/Delete credentials in the OS secret store (macOS Keychain on macOS, secret-service via D-Bus on Linux, credman on Windows). Uses `zalando/go-keyring` v0.2.8 (Go 1.18+, no CGO).
- New `bitrise-build-cache auth` command group: `set`, `get` (masked last-4), `clear`.
- New `configcommon.ResolveAuthConfig(envs)` — tries keychain first, falls back to existing env-var read on any error (ErrNotFound, missing D-Bus, partial creds, etc). The original `ReadAuthConfigFromEnvironments` is kept for callers that explicitly want env-only behavior.
- All 12 production auth-read call sites migrated to `ResolveAuthConfig`. Tests are unchanged where they exercise env-only logic.

## Why

Closes [ACI-5028](https://bitrise.atlassian.net/browse/ACI-5028). Customers shouldn't have to keep `BITRISE_BUILD_CACHE_AUTH_TOKEN` in their shell rc files. CI keeps working unchanged because keychain errors transparently fall through to env.

## What's intentionally NOT in this PR

- **Migration prompt** (env → keychain) is deferred to [ACI-5027 (A1 — `activate --interactive` wizard)](https://bitrise.atlassian.net/browse/ACI-5027), which is the natural surface for an interactive import flow. Per the spike-discussion decision (Q4 = c, only inside the wizard).
- Masked stdin prompt on `auth set`. Currently `--token` and `--workspace-id` are required flags. The masked-prompt UX belongs to the wizard ticket.

## Test plan

- [x] `go build ./...` clean
- [x] `make lint` — 0 issues
- [x] `make test-unit` — all packages green
- [x] New unit tests:
  - `internal/auth/keychain/keychain_test.go` — round-trip, ErrNotFound, replace-value, clear, backend-error wrapping.
  - `internal/config/common/auth_test.go` — `ResolveAuthConfig` with fake loader: keychain hit, ErrNotFound fallthrough, backend-error fallthrough, empty-creds fallthrough, partial-creds fallthrough.
- [x] Smoke test on macOS: `auth set` → `auth get` shows workspace + `****<last4>` → `auth clear` → `auth get` reports "no credentials".

[ACI-5028]: https://bitrise.atlassian.net/browse/ACI-5028?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cmd/auth/auth.go
+++ b/cmd/auth/auth.go
@@ -1,14 +1,22 @@
 package auth
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"strings"
 
 	"github.com/bitrise-io/go-utils/v2/log"
 	"github.com/spf13/cobra"
 
 	"github.com/bitrise-io/bitrise-build-cache-cli/v2/cmd/common"
 	"github.com/bitrise-io/bitrise-build-cache-cli/v2/internal/auth/keychain"
+	configcommon "github.com/bitrise-io/bitrise-build-cache-cli/v2/internal/config/common"
+	multiplatformconfig "github.com/bitrise-io/bitrise-build-cache-cli/v2/internal/config/multiplatform"
+	"github.com/bitrise-io/bitrise-build-cache-cli/v2/internal/utils"
 )
 
 // nolint:gochecknoglobals
@@ -51,36 +59,236 @@ var authSetCmd = &cobra.Command{
 		}
 
 		logger.TInfof("✅ Credentials saved to the OS keychain")
+
+		switch scrubbed, err := scrubDiskCredentials(); {
+		case err != nil:
+			logger.Warnf("Saved to keychain, but could not strip plain-text credentials from disk: %v", err)
+			logger.Warnf("Run `bitrise-build-cache auth get` to audit remaining sources.")
+		case len(scrubbed) > 0:
+			logger.TInfof("Scrubbed plain-text credentials from %s", strings.Join(scrubbed, ", "))
+		}
+
 		logger.Infof("You can now remove BITRISE_BUILD_CACHE_AUTH_TOKEN + BITRISE_BUILD_CACHE_WORKSPACE_ID from your shell rc files.")
+		logger.Infof("If you have running Gradle daemons, stop them so the new token is picked up: `./gradlew --stop`.")
 
 		return nil
 	},
 }
 
+func scrubDiskCredentials() ([]string, error) {
+	osProxy := utils.DefaultOsProxy{}
+
+	cfg, err := multiplatformconfig.ReadConfig(osProxy, utils.DefaultDecoderFactory{})
+	switch {
+	case errors.Is(err, fs.ErrNotExist):
+		return nil, nil
+	case err != nil:
+		return nil, fmt.Errorf("read multiplatform config: %w", err)
+	}
+
+	if cfg.AuthConfig.AuthToken == "" && cfg.AuthConfig.WorkspaceID == "" {
+		return nil, nil
+	}
+
+	cfg.AuthConfig = configcommon.CacheAuthConfig{}
+	if err := cfg.Save(osProxy, utils.DefaultEncoderFactory{}); err != nil {
+		return nil, fmt.Errorf("save scrubbed multiplatform config: %w", err)
+	}
+
+	return []string{"~/.bitrise/analytics/multiplatform/config.json"}, nil
+}
+
 // nolint:gochecknoglobals
 var authGetCmd = &cobra.Command{
 	Use:          "get",
-	Short:        "Show whether Bitrise Build Cache credentials are stored in the OS keychain",
+	Short:        "Show Bitrise Build Cache credentials discovered across all known sources",
+	Long:         "Lists credentials found in the OS keychain, the multiplatform analytics config on disk, and the BITRISE_BUILD_CACHE_AUTH_TOKEN / BITRISE_BUILD_CACHE_WORKSPACE_ID / BITRISEIO_BITRISE_SERVICES_ACCESS_TOKEN env vars. Use this to audit where your credentials live and to migrate them to the OS keychain.",
 	SilenceUsage: true,
 	RunE: func(_ *cobra.Command, _ []string) error {
 		logger := log.NewLogger(log.WithDebugLog(common.IsDebugLogMode))
 
-		creds, err := keychain.New().Load()
-		switch {
-		case errors.Is(err, keychain.ErrNotFound):
-			logger.Warnf("No Bitrise Build Cache credentials stored in the OS keychain.")
-			logger.Infof("Run `bitrise-build-cache auth set --token <token> --workspace-id <id>` to store them, or rely on env vars.")
+		sources := credSources(utils.AllEnvs())
 
-			return nil
-		case err != nil:
-			return fmt.Errorf("read credentials from keychain: %w", err)
+		var keychainPopulated, foundElsewhere bool
+		for i, s := range sources {
+			populated := printSource(logger, s)
+			switch {
+			case i == 0:
+				keychainPopulated = populated
+			case populated:
+				foundElsewhere = true
+			}
 		}
 
-		logger.TInfof("Workspace ID: %s", creds.WorkspaceID)
-		logger.TInfof("Auth token:   %s", maskToken(creds.AuthToken))
+		if !keychainPopulated && foundElsewhere {
+			logger.Println()
+			logger.Infof("Credentials exist outside the OS keychain — migrate with:")
+			logger.Infof("  bitrise-build-cache auth set --token <token> --workspace-id <workspace-id>")
+			logger.Infof("`auth set` scrubs the multiplatform config in place; xcelerate and ccache configs are re-written cleanly on the next `activate xcode` / `activate c++`.")
+		}
 
 		return nil
 	},
+}
+
+type credSourceState int
+
+const (
+	sourceAbsent credSourceState = iota
+	sourcePartial
+	sourceReadError
+	sourcePopulated
+	sourcePopulatedTokenOnly
+)
+
+type credAudit struct {
+	state       credSourceState
+	workspaceID string
+	authToken   string
+	note        string
+	err         error
+}
+
+type credSource struct {
+	label    string
+	location string
+	probe    func() credAudit
+}
+
+// First entry MUST be the keychain — RunE treats sources[0] as the target.
+func credSources(envs map[string]string) []credSource {
+	return []credSource{
+		{"OS keychain", "<system keychain>", probeKeychain},
+		{"Multiplatform config", "~/.bitrise/analytics/multiplatform/config.json", probeMultiplatform},
+		{"Xcelerate config", "~/.bitrise-xcelerate/config.json", probeRawConfig("~/.bitrise-xcelerate/config.json")},
+		{"Ccache config", "~/.bitrise/cache/ccache/config.json", probeRawConfig("~/.bitrise/cache/ccache/config.json")},
+		{"Env vars (BITRISE_BUILD_CACHE_AUTH_TOKEN + BITRISE_BUILD_CACHE_WORKSPACE_ID)", "process env", func() credAudit { return probeEnvVars(envs) }},
+		{"CI JWT (BITRISEIO_BITRISE_SERVICES_ACCESS_TOKEN)", "process env", func() credAudit { return probeJWT(envs) }},
+	}
+}
+
+func printSource(logger log.Logger, s credSource) bool {
+	if s.location != "" {
+		logger.TInfof("%s (%s):", s.label, s.location)
+	} else {
+		logger.TInfof("%s:", s.label)
+	}
+
+	a := s.probe()
+	switch a.state {
+	case sourceAbsent:
+		if a.note != "" {
+			logger.Infof("  %s", a.note)
+		} else {
+			logger.Infof("  not configured")
+		}
+
+		return false
+	case sourcePartial:
+		logger.Warnf("  partial: %s", a.note)
+
+		return false
+	case sourceReadError:
+		logger.Errorf("  read failed: %v", a.err)
+
+		return false
+	case sourcePopulated:
+		logger.Infof("  Workspace ID: %s", a.workspaceID)
+		logger.Infof("  Auth token:   %s", maskToken(a.authToken))
+
+		return true
+	case sourcePopulatedTokenOnly:
+		logger.Infof("  set (%s)", maskToken(a.authToken))
+
+		return true
+	}
+
+	return false
+}
+
+func probeKeychain() credAudit {
+	creds, err := keychain.New().Load()
+	switch {
+	case errors.Is(err, keychain.ErrNotFound):
+		return credAudit{state: sourceAbsent}
+	case err != nil:
+		return credAudit{state: sourceReadError, err: err}
+	}
+
+	return credAudit{state: sourcePopulated, workspaceID: creds.WorkspaceID, authToken: creds.AuthToken}
+}
+
+func probeMultiplatform() credAudit {
+	cfg, err := multiplatformconfig.ReadConfig(utils.DefaultOsProxy{}, utils.DefaultDecoderFactory{})
+	switch {
+	case errors.Is(err, fs.ErrNotExist):
+		return credAudit{state: sourceAbsent, note: "not present"}
+	case err != nil:
+		return credAudit{state: sourceReadError, err: err}
+	case cfg.AuthConfig.AuthToken == "":
+		return credAudit{state: sourceAbsent, note: "present but no credentials"}
+	}
+
+	return credAudit{state: sourcePopulated, workspaceID: cfg.AuthConfig.WorkspaceID, authToken: cfg.AuthConfig.AuthToken}
+}
+
+// probeRawConfig reads the file directly — the per-tool ReadConfig overlays multiplatform credentials, which would mask the actual on-disk content this audit needs to see.
+func probeRawConfig(displayPath string) func() credAudit {
+	return func() credAudit {
+		home, err := utils.DefaultOsProxy{}.UserHomeDir()
+		if err != nil {
+			return credAudit{state: sourceReadError, err: fmt.Errorf("resolve home dir: %w", err)}
+		}
+
+		fullPath := filepath.Join(home, strings.TrimPrefix(displayPath, "~/"))
+
+		body, err := os.ReadFile(fullPath) //nolint:gosec // path composed from home + constant
+		switch {
+		case errors.Is(err, fs.ErrNotExist):
+			return credAudit{state: sourceAbsent, note: "not present"}
+		case err != nil:
+			return credAudit{state: sourceReadError, err: err}
+		}
+
+		var raw struct {
+			AuthConfig struct {
+				AuthToken   string `json:"authToken"`
+				WorkspaceID string `json:"workspaceID"`
+			} `json:"authConfig"`
+		}
+		if err := json.Unmarshal(body, &raw); err != nil {
+			return credAudit{state: sourceReadError, err: fmt.Errorf("decode: %w", err)}
+		}
+
+		if raw.AuthConfig.AuthToken == "" {
+			return credAudit{state: sourceAbsent, note: "present but no embedded credentials"}
+		}
+
+		return credAudit{state: sourcePopulated, workspaceID: raw.AuthConfig.WorkspaceID, authToken: raw.AuthConfig.AuthToken}
+	}
+}
+
+func probeEnvVars(envs map[string]string) credAudit {
+	tok := envs["BITRISE_BUILD_CACHE_AUTH_TOKEN"]
+	ws := envs["BITRISE_BUILD_CACHE_WORKSPACE_ID"]
+
+	switch {
+	case tok != "" && ws != "":
+		return credAudit{state: sourcePopulated, workspaceID: ws, authToken: tok}
+	case tok != "" || ws != "":
+		return credAudit{state: sourcePartial, note: "only one of AUTH_TOKEN / WORKSPACE_ID is set"}
+	}
+
+	return credAudit{state: sourceAbsent, note: "not set"}
+}
+
+func probeJWT(envs map[string]string) credAudit {
+	jwt := envs["BITRISEIO_BITRISE_SERVICES_ACCESS_TOKEN"]
+	if jwt == "" {
+		return credAudit{state: sourceAbsent, note: "not set"}
+	}
+
+	return credAudit{state: sourcePopulatedTokenOnly, authToken: jwt}
 }
 
 // nolint:gochecknoglobals
@@ -96,6 +304,29 @@ var authClearCmd = &cobra.Command{
 		}
 
 		logger.TInfof("✅ Credentials removed from the OS keychain")
+
+		return nil
+	},
+}
+
+// nolint:gochecknoglobals
+var authTokenCmd = &cobra.Command{
+	Use:           "token",
+	Short:         "Resolve and print the Bitrise Build Cache auth token to stdout",
+	Long:          "Resolves the auth token via the same precedence chain as the rest of the CLI (env vars → OS keychain → multiplatform analytics config) and prints it to stdout. Intended for build-time consumers (Gradle init script, future Bazel workspace_status_command) that need the resolved token without baking it into a config file. On failure exits non-zero with a short one-line message on stderr (no cobra Error: prefix) — callers framing the wrapper script own the wording.",
+	SilenceUsage:  true,
+	SilenceErrors: true,
+	RunE: func(cmd *cobra.Command, _ []string) error {
+		cfg, err := configcommon.ResolveAuthConfig(utils.AllEnvs())
+		if err != nil {
+			_, _ = fmt.Fprintln(cmd.ErrOrStderr(), err.Error())
+
+			return fmt.Errorf("resolve auth config: %w", err)
+		}
+
+		if _, err := fmt.Fprintln(cmd.OutOrStdout(), cfg.AuthToken); err != nil {
+			return fmt.Errorf("write auth token: %w", err)
+		}
 
 		return nil
 	},
@@ -119,6 +350,7 @@ func init() {
 	authCmd.AddCommand(authSetCmd)
 	authCmd.AddCommand(authGetCmd)
 	authCmd.AddCommand(authClearCmd)
+	authCmd.AddCommand(authTokenCmd)
 
 	common.RootCmd.AddCommand(authCmd)
 }
--- a/cmd/auth/auth.go
+++ b/cmd/auth/auth.go
@@ -15,7 +15,7 @@ import (
 var authCmd = &cobra.Command{
 	Use:          "auth",
 	Short:        "Manage Bitrise Build Cache credentials stored in the OS keychain",
-	Long:         `Manage Bitrise Build Cache credentials stored in the OS keychain (macOS Keychain, Linux secret-service). Credentials persisted here are used in preference to the BITRISE_BUILD_CACHE_AUTH_TOKEN / BITRISE_BUILD_CACHE_WORKSPACE_ID env vars.`,
+	Long:         `Manage Bitrise Build Cache credentials stored in the OS keychain (macOS Keychain, Linux secret-service). Stored credentials are used when BITRISE_BUILD_CACHE_AUTH_TOKEN / BITRISE_BUILD_CACHE_WORKSPACE_ID (or BITRISEIO_BITRISE_SERVICES_ACCESS_TOKEN on Bitrise CI) are not set — env vars take precedence so you can override the stored credentials for a single run.`,
 	SilenceUsage: true,
 }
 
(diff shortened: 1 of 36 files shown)


zsolt-marta-bitrise:
[//]: # ([skip ci])



## Summary
Implements OS keychain credential storage with new `bitrise-build-cache auth` command group (set/list/clear/token). Migrates all 12 production auth-read call sites to `ResolveAuthConfig()` which tries keychain first, falls back to env vars on any error. Adds keychain backend abstraction, multiplatform config reader registration, and gradle init script token resolution for local dev.


## Walkthrough
| File | Summary |
|------|---------|
| internal/auth/keychain/keychain.go | New package implementing OS keychain Get/Set/Delete via zalando/go-keyring (macOS Keychain, Linux D-Bus, Windows credman) |
| cmd/auth/auth.go | New auth command group: set (store creds + scrub disk), list (audit all sources), clear (remove from keychain), token (resolve + print) |
| internal/config/common/auth.go | New ResolveAuthConfig() tries env vars → keychain → multiplatform config with fallback chain; ReadAuthConfigFromEnvironments kept for env-only callers |
| internal/config/{gradle,bazel,xcelerate,ccache}/ | Migrated auth resolution to ResolveAuthConfig; updated error messages from 'read from environments' to 'resolve auth config' |
| cmd/{gradle,xcode}/ | Migrated auth resolution to ResolveAuthConfig with proper error handling; added CLIPath init for gradle |
| internal/config/gradle/initd.gradle.kts.gotemplate | Added BitriseAuthTokenSource class + bitriseAuthToken() extension for local dev token resolution via CLI |
| internal/config/multiplatform/register.go, internal/paths/paths.go | Added multiplatform reader registration and gradle init script path resolution helpers |
| pkg/{file,reactnative}/, main.go | Migrated remaining auth calls to ResolveAuthConfig; registered auth command in CLI |
| go.mod, go.sum, *_test.go | Added zalando/go-keyring v0.2.8 dependency; comprehensive unit tests for keychain, ResolveAuthConfig, and command integration |

internal/config/gradle/activate.go
```diff
@@ -25,7 +25,7 @@ func Activate(
 	updater GradlePropertiesUpdater,
 	params ActivateGradleParams,
 ) error {
-	authConfig, _ := configcommon.ReadAuthConfigFromEnvironments(envProvider)
+	authConfig, _ := configcommon.ResolveAuthConfig(envProvider)
```
zsolt-marta-bitrise: [bitrise-plugin-ai-reviewer]: internal/config/gradle/activate.go:28:0:8b2bbb22

[//]: # ([skip ci])


**🐛 Bug**

Error from ResolveAuthConfig is silently discarded. This is inconsistent with the error handling in activate_for_gradle_params.go (line 81-84) where the error is properly checked and returned. If auth resolution fails (e.g., missing credentials), the code will continue with empty authConfig instead of propagating the error. While BenchmarkPhaseClient is defensive about empty credentials, this is a logic error that should be caught and handled properly.

🔄 Suggestion:
```suggestion
	authConfig, err := configcommon.ResolveAuthConfig(envProvider)
	if err != nil {
		return fmt.Errorf("resolve auth config: %w", err)
	}
```

zsolt-marta-bitrise (commented):

zsolt-marta-bitrise (commented):

cmd/gradle/enable_for_gradle.go
```diff
@@ -82,7 +82,7 @@ func EnableForGradleCmdFn(logger log.Logger, gradleHomePath string, envProvider
 	activateGradleParams.Analytics.Enabled = paramIsGradleMetricsEnabled
 	activateGradleParams.TestDistro.Enabled = false
 
-	authConfig, _ := configcommon.ReadAuthConfigFromEnvironments(envProvider)
+	authConfig, _ := configcommon.ResolveAuthConfig(envProvider)
```
zsolt-marta-bitrise: [bitrise-plugin-ai-reviewer]: cmd/gradle/enable_for_gradle.go:85:0:8b2bbb22

[//]: # ([skip ci])


**🐛 Bug**

Error from ResolveAuthConfig is silently discarded instead of being checked and returned. This is inconsistent with the error handling in restoreGradleConfigurationCache.go, saveGradleConfigurationCache.go, and activate_for_gradle_params.go, which all properly check and return the error. If ResolveAuthConfig fails (e.g., keychain error, missing credentials), the error is silently ignored and an empty authConfig is used, which could lead to authentication failures downstream.

🔄 Suggestion:
```suggestion
	authConfig, err := configcommon.ResolveAuthConfig(envProvider)
	if err != nil {
		return fmt.Errorf("resolve auth config: %w", err)
	}
```

internal/config/gradle/activate.go
Badlazzor: Fixed in 2906e68 — the outer ResolveAuthConfig error is now wrapped and returned via `ErrFmtReadAuthConfig`, matching the activate_for_gradle_params.go pattern.

cmd/gradle/enable_for_gradle.go
Badlazzor: Fixed in 0e20bec — same wrap+return pattern applied at enable_for_gradle.go:85.

zsolt-marta-bitrise (commented):

internal/config/xcelerate/activate.go
```diff
@@ -63,7 +63,7 @@ func Activate(
 	overrideActivateXcodeParamsFromExistingConfig(
 		logger, osProxy, &activateXcodeParams, decoderFactory, envs)
 
-	authConfig, _ := configcommon.ReadAuthConfigFromEnvironments(envs)
+	authConfig, _ := configcommon.ResolveAuthConfig(envs)
```
zsolt-marta-bitrise: [bitrise-plugin-ai-reviewer]: internal/config/xcelerate/activate.go:66:0:8b2bbb22

[//]: # ([skip ci])


**🐛 Bug**

Error from ResolveAuthConfig is silently discarded instead of being checked and returned like in gradle/activate.go. If ResolveAuthConfig fails, authConfig will be a zero-value struct, which could cause unexpected behavior when passed to NewBenchmarkPhaseClient.

🔄 Suggestion:
```suggestion
	authConfig, err := configcommon.ResolveAuthConfig(envs)
	if err != nil {
		return fmt.Errorf("read auth config: %w", err)
	}
```

Badlazzor: Fixed in ede806d — same wrap+return pattern applied at xcelerate/activate.go:66.

zsolt-marta-bitrise (commented):

cmd/xcode/deleteXcodeDerivedData.go
```diff
@@ -70,7 +70,7 @@ func deleteXcodeDerivedDataCmdFn(ctx context.Context,
 	commandFunc func(string, ...string) (string, error),
 ) error {
 	logger.Infof("(i) Check Auth Config")
-	authConfig, err := configcommon.ReadAuthConfigFromEnvironments(envProvider)
+	authConfig, err := configcommon.ResolveAuthConfig(envProvider)
 	if err != nil {
 		return fmt.Errorf("read auth config from environments: %w", err)
```
zsolt-marta-bitrise: [bitrise-plugin-ai-reviewer]: cmd/xcode/deleteXcodeDerivedData.go:75:0:e8f43e15

[//]: # ([skip ci])


**📋 Guideline**

The error message says "read auth config from environments" but the function being called is `ResolveAuthConfig`, which tries the keychain first and falls back to environment variables. The error message is misleading and doesn't accurately describe what the function does.

🔄 Suggestion:
```suggestion
		return fmt.Errorf("resolve auth config: %w", err)
```

cmd/xcode/saveXcodeDerivedDataFiles.go
```diff
@@ -51,7 +51,7 @@ var saveXcodeDerivedDataFilesCmd = &cobra.Command{
 
 		logger.Infof("(i) Check Auth Config")
 		allEnvs := utils.AllEnvs()
-		authConfig, err := configcommon.ReadAuthConfigFromEnvironments(allEnvs)
+		authConfig, err := configcommon.ResolveAuthConfig(allEnvs)
 		if err != nil {
 			return fmt.Errorf("read auth config from environments: %w", err)
```
zsolt-marta-bitrise: [bitrise-plugin-ai-reviewer]: cmd/xcode/saveXcodeDerivedDataFiles.go:56:0:96951f4e

[//]: # ([skip ci])


**📋 Guideline**

The error message says "read auth config from environments" but the function being called is `ResolveAuthConfig`, which tries the keychain first and falls back to environment variables. The error message is misleading and doesn't accurately describe what the function does.

🔄 Suggestion:
```suggestion
			return fmt.Errorf("resolve auth config: %w", err)
```

cmd/xcode/deleteXcodeDerivedData.go
Badlazzor: Fixed in 0b818d4 — wrap reworded to `"resolve auth config: %w"` so it covers both keychain and env-var lookup paths. Applied across all 6 sites that used the misleading "from environments" wording (3 cmd/xcode, 2 cmd/gradle, 1 pkg/file).

cmd/xcode/saveXcodeDerivedDataFiles.go
Badlazzor: Fixed in 0b818d4 — wrap reworded to `"resolve auth config: %w"` so it covers both keychain and env-var lookup paths. Applied across all 6 sites that used the misleading "from environments" wording (3 cmd/xcode, 2 cmd/gradle, 1 pkg/file).

zsolt-marta-bitrise (commented):

internal/config/bazel/activate_bazel_params.go
```diff
@@ -82,7 +82,7 @@ func (params ActivateBazelParams) commonTemplateInventory(
 
 	// Required configs
 	logger.Infof("(i) Check Auth Config")
-	authConfig, err := common.ReadAuthConfigFromEnvironments(envs)
+	authConfig, err := common.ResolveAuthConfig(envs)
 	if err != nil {
 		return CommonTemplateInventory{},
 			fmt.Errorf("read auth config from environment variables: %w", err)
```
zsolt-marta-bitrise: [bitrise-plugin-ai-reviewer]: internal/config/bazel/activate_bazel_params.go:88:0:b6e9c0ed

[//]: # ([skip ci])


**📋 Guideline**

Misleading error message. The function `ResolveAuthConfig` tries the keychain first and falls back to environment variables on any error. The error message should reflect this behavior, not just mention environment variables.

🔄 Suggestion:
```suggestion
			fmt.Errorf("resolve auth config: %w", err)
```

Badlazzor: Fixed in 0445511 — the same wrap rename swept through `internal/config` too. `activate_bazel_params.go:88`, `internal/config/xcelerate/activate.go` and the `gradle.ErrFmtReadAuthConfig` const all read `resolve auth config: %w` now, plus the two tests that asserted on the old string.

zsolt-marta-bitrise (commented):

cmd/auth/auth.go
```diff
@@ -0,0 +1,116 @@
...
+	setToken       string
+	setWorkspaceID string
+)
+
+// nolint:gochecknoglobals
+var authSetCmd = &cobra.Command{
+	Use:          "set",
+	Short:        "Store Bitrise Build Cache credentials in the OS keychain",
+	SilenceUsage: true,
+	RunE: func(_ *cobra.Command, _ []string) error {
```
zsolt-marta-bitrise, line 45: [bitrise-plugin-ai-reviewer]: cmd/auth/auth.go:33:0:8b2bbb22

[//]: # ([skip ci])


**🐛 Bug**

Missing validation of flag values. Cobra's MarkFlagRequired only checks if the flag is provided, not if it's empty. A user could pass `--token ""` and `--workspace-id ""` and the command would save empty credentials to the keychain without error.

🔄 Suggestion:
```suggestion
	RunE: func(_ *cobra.Command, _ []string) error {
		logger := log.NewLogger(log.WithDebugLog(common.IsDebugLogMode))

		if setToken == "" {
			return errors.New("--token flag cannot be empty")
		}
		if setWorkspaceID == "" {
			return errors.New("--workspace-id flag cannot be empty")
		}

		kc := keychain.New()
		if err := kc.Save(keychain.Credentials{
			AuthToken:   setToken,
			WorkspaceID: setWorkspaceID,
		}); err != nil {
			return fmt.Errorf("save credentials to keychain: %w", err)
		}
```

cmd/auth/auth.go
```diff
@@ -0,0 +1,124 @@
...
+		kc := keychain.New()
+		if err := kc.Save(keychain.Credentials{
+			AuthToken:   setToken,
+			WorkspaceID: setWorkspaceID,
+		}); err != nil {
+			return fmt.Errorf("save credentials to keychain: %w", err)
+		}
+
+		logger.TInfof("✅ Credentials saved to the OS keychain")
+		logger.Infof("You can now remove BITRISE_BUILD_CACHE_AUTH_TOKEN + BITRISE_BUILD_CACHE_WORKSPACE_ID from your shell rc files.")
```
zsolt-marta-bitrise, line 88: Do we want to remove these from the config files / sidecars too in this case (if we have that)? For security conscious users it might be helpful

However, not sure how that would improve security here 🤔 If we still have tokens in the Gradle init script / xcc config file, what would they save by putting the token to the keychain

cmd/auth/auth.go
```diff
@@ -0,0 +1,124 @@
...
+
+		return nil
+	},
+}
+
+// nolint:gochecknoglobals
+var authGetCmd = &cobra.Command{
+	Use:          "get",
+	Short:        "Show whether Bitrise Build Cache credentials are stored in the OS keychain",
```
zsolt-marta-bitrise: This seems like a generic command `auth get`, with no mention that this is keychain specifc.  In case we have set credentials in activated tools' sidecards/configs we'd need to list them to be able to assess where they have set a token.

Also, I'd still print here what we have currently in the envs, so the result of the ResolveAuthConfig method with source specified. That can help users clean up their config to remove envs from everywhere if they migrate to keychain

zsolt-marta-bitrise: Alternatively, we can scope the root command to one level down: `build-cache-cli auth keychain get/set/remove`

cmd/auth/auth.go
```diff
@@ -0,0 +1,124 @@
...
+		logger.TInfof("Auth token:   %s", maskToken(creds.AuthToken))
+
+		return nil
+	},
+}
+
+// nolint:gochecknoglobals
+var authClearCmd = &cobra.Command{
+	Use:          "clear",
+	Short:        "Remove Bitrise Build Cache credentials from the OS keychain",
```
zsolt-marta-bitrise, line 472: Same as above, include config files / sidecars

cmd/auth/auth.go
```diff
@@ -0,0 +1,356 @@
...
+
+// nolint:gochecknoglobals
+var authSetCmd = &cobra.Command{
+	Use:          "set",
+	Short:        "Store Bitrise Build Cache credentials in the OS keychain",
+	SilenceUsage: true,
+	RunE: func(_ *cobra.Command, _ []string) error {
+		logger := log.NewLogger(log.WithDebugLog(common.IsDebugLogMode))
+
+		switch {
```
zsolt-marta-bitrise, line 51: [bitrise-plugin-ai-reviewer]: cmd/auth/auth.go:44:0:8a545e1e

[//]: # ([skip ci])


**🐛 Bug**

The validation checks for empty strings but does not trim whitespace. A user could pass `--token "   "` (spaces only) which would pass validation but result in invalid credentials being saved to the keychain.

🔄 Suggestion:
```suggestion
		setToken = strings.TrimSpace(setToken)
		setWorkspaceID = strings.TrimSpace(setWorkspaceID)

		switch {
		case setToken == "" && setWorkspaceID == "":
			return errors.New("--token and --workspace-id are required and must not be empty")
		case setToken == "":
			return errors.New("--token is required and must not be empty")
		case setWorkspaceID == "":
			return errors.New("--workspace-id is required and must not be empty")
		}
```

zsolt-marta-bitrise (commented):

pkg/reactnative/activator.go
```diff
@@ -348,7 +348,7 @@ func (s *storageHelperStarter) start() error {
 }
 
 func saveMultiplatformConfig(debugLogging bool) error {
-	authConfig, err := configcommon.ReadAuthConfigFromEnvironments(utils.AllEnvs())
+	authConfig, err := configcommon.ResolveAuthConfig(utils.AllEnvs())
 	if err != nil {
 		return fmt.Errorf("read auth config for multiplatform analytics: %w", err)
```
zsolt-marta-bitrise: [bitrise-plugin-ai-reviewer]: pkg/reactnative/activator.go:353:0:3752dc54

[//]: # ([skip ci])


**📋 Guideline**

The error message says "read auth config" but the function now calls ResolveAuthConfig which tries keychain first, then falls back to environment variables. The message should reflect this behavior change, similar to the update made in pkg/file/helper.go line 157.

🔄 Suggestion:
```suggestion
		return fmt.Errorf("resolve auth config for multiplatform analytics: %w", err)
```

zsolt-marta-bitrise (commented):

cmd/auth/auth.go
Badlazzor: - **Gradle init.d**: this PR rewrites the script so the token is resolved at build time via a Gradle ValueSource that shells out to `bitrise-build-cache auth token` (commits `d36a711` → `78e8a27`). No literal token in the file.
- **xcelerate config**: `Save` already zeroes `AuthConfig` before writing, so the next `activate xcode` rewrites cleanly.
- **ccache config**: `AuthConfig` is `json:\"-\"` (never persisted) and the in-memory value is sourced from multiplatform at read time.
- **multiplatform config**: `scrubDiskCredentials` in this commit (`2e63136`) zeroes the AuthConfig in place after a successful keychain save.
- **bazel `~/.bazelrc`**: plain-text remains; tracked separately as [ACI-5116](https://bitrise.atlassian.net/browse/ACI-5116) — same dynamic-resolution approach (workspace_status_command shell-out is the leading candidate).

So after `auth set` succeeds, gradle / xcelerate / ccache / multiplatform stop leaking either immediately or on the next `activate`. Bazel is the last remaining source until ACI-5116 lands.

cmd/auth/auth.go
Badlazzor: Good point — landed in `2e63136` and refined to a data-driven audit table in `78e8a27`. `auth get` now prints every known source with state + a migration hint when creds live outside the keychain:

- OS keychain
- Multiplatform config (`~/.bitrise/analytics/multiplatform/config.json`)
- Xcelerate config (`~/.bitrise-xcelerate/config.json`) — raw read so the multiplatform-overlay in `ReadConfig` can't mask what's literally on disk
- Ccache config (`~/.bitrise/cache/ccache/config.json`) — same raw-read treatment
- `BITRISE_BUILD_CACHE_AUTH_TOKEN` + `BITRISE_BUILD_CACHE_WORKSPACE_ID` env vars (with a partial-set warning)
- `BITRISEIO_BITRISE_SERVICES_ACCESS_TOKEN` (CI JWT) env var

Renaming the subtree to `auth keychain get/set/...` was on the table, but since the audit now covers every source the generic naming reads correctly. If you'd still prefer the scoped subcommand, easy to add.

cmd/auth/auth.go
Badlazzor: Same answer as the thread above on line 74: `scrubDiskCredentials` handles multiplatform on `auth set`; xcelerate / ccache self-clean on next `activate` (they zero `AuthConfig` on Save); gradle init.d no longer carries a literal token after this PR's ValueSource shell-out; bazel pending [ACI-5116](https://bitrise.atlassian.net/browse/ACI-5116).

cmd/auth/auth.go
```diff
@@ -0,0 +1,359 @@
...
+	RunE: func(_ *cobra.Command, _ []string) error {
+		logger := log.NewLogger(log.WithDebugLog(common.IsDebugLogMode))
+
+		sources := credSources(utils.AllEnvs())
+
+		var keychainPopulated, foundElsewhere bool
+		for i, s := range sources {
+			populated := printSource(logger, s)
+			switch {
+			case i == 0:
```
zsolt-marta-bitrise: Please dont depend having the keychain at position 0, rather check ID

cmd/auth/auth.go
```diff
@@ -0,0 +1,359 @@
...
+	err         error
+}
+
+type credSource struct {
+	label    string
+	location string
+	probe    func() credAudit
+}
+
+// First entry MUST be the keychain — RunE treats sources[0] as the target.
```
zsolt-marta-bitrise: I wouldn't do this, fragile code

cmd/auth/auth.go
```diff
@@ -0,0 +1,359 @@
...
+func credSources(envs map[string]string) []credSource {
+	return []credSource{
+		{"OS keychain", "<system keychain>", probeKeychain},
+		{"Multiplatform config", "~/.bitrise/analytics/multiplatform/config.json", probeMultiplatform},
```
zsolt-marta-bitrise: Can we depend on `paths` package for these? That way the path references stay the same place

cmd/auth/auth.go
```diff
@@ -0,0 +1,359 @@
...
+		{"Xcelerate config", "~/.bitrise-xcelerate/config.json", probeRawConfig("~/.bitrise-xcelerate/config.json")},
+		{"Ccache config", "~/.bitrise/cache/ccache/config.json", probeRawConfig("~/.bitrise/cache/ccache/config.json")},
+		{"Env vars (BITRISE_BUILD_CACHE_AUTH_TOKEN + BITRISE_BUILD_CACHE_WORKSPACE_ID)", "process env", func() credAudit { return probeEnvVars(envs) }},
+		{"CI JWT (BITRISEIO_BITRISE_SERVICES_ACCESS_TOKEN)", "process env", func() credAudit { return probeJWT(envs) }},
+	}
+}
+
+func printSource(logger log.Logger, s credSource) bool {
```
zsolt-marta-bitrise: This is great information, but I'd only log by default the ones that are not absent (so they have a token). The user will probably never activate tools outside with what they work with + this only applies if they have activated before the release of this PR. For debug we can print all states though

cmd/auth/auth.go
```diff
@@ -0,0 +1,359 @@
...
+		}
+
+		return nil
+	},
+}
+
+func maskToken(token string) string {
+	const tailLen = 4
+	if len(token) <= tailLen {
+		return "(present, length too short to mask)"
```
zsolt-marta-bitrise, line 513: Spent a bit time to think about this, but probably 4 characters and only locally ad-hoc should be fine

cmd/auth/auth.go
```diff
@@ -0,0 +1,359 @@
...
+	if err := cfg.Save(osProxy, utils.DefaultEncoderFactory{}); err != nil {
+		return nil, fmt.Errorf("save scrubbed multiplatform config: %w", err)
+	}
+
+	return []string{"~/.bitrise/analytics/multiplatform/config.json"}, nil
+}
+
+// nolint:gochecknoglobals
+var authGetCmd = &cobra.Command{
+	Use:          "get",
```
zsolt-marta-bitrise: Probably we should call this `list` now that it discovers and lists all sources + we have a resolve command

cmd/auth/auth.go
```diff
@@ -0,0 +1,359 @@
...
+	configcommon "github.com/bitrise-io/bitrise-build-cache-cli/v2/internal/config/common"
+	multiplatformconfig "github.com/bitrise-io/bitrise-build-cache-cli/v2/internal/config/multiplatform"
+	"github.com/bitrise-io/bitrise-build-cache-cli/v2/internal/utils"
+)
+
+// nolint:gochecknoglobals
+var authCmd = &cobra.Command{
+	Use:          "auth",
+	Short:        "Manage Bitrise Build Cache credentials stored in the OS keychain",
+	Long:         `Manage Bitrise Build Cache credentials stored in the OS keychain (macOS Keychain, Linux secret-service). Stored credentials are used when BITRISE_BUILD_CACHE_AUTH_TOKEN / BITRISE_BUILD_CACHE_WORKSPACE_ID (or BITRISEIO_BITRISE_SERVICES_ACCESS_TOKEN on Bitrise CI) are not set — env vars take precedence so you can override the stored credentials for a single run.`,
```
zsolt-marta-bitrise, line 30: is the linux implemented?

cmd/auth/auth.go
zsolt-marta-bitrise: I see gradle would be a bit risky to migrate because of the different init.d format. But what about xcode? We could scrub that too

cmd/auth/auth.go
Badlazzor: yes, see keychain.go

cmd/auth/auth.go
Badlazzor: Done in 4ee7a18 — scrub extended to xcelerate + ccache `config.json` `authConfig` blocks and gradle init.kts `authToken` literals blanked to `""`. Per-source hint points at the right `activate <tool>` to regenerate cleanly.

cmd/auth/auth.go
Badlazzor: Kept flat — `list` (renamed in 4ee7a18) audits ALL sources, not just keychain, so `auth keychain list` would mislead. Happy to revisit if we ever add a non-keychain `set`.

cmd/auth/auth.go
Badlazzor: Covered by the scrub extension in 4ee7a18 — xcelerate/ccache configs + gradle init.kts.

cmd/auth/auth.go
Badlazzor: Done in 4ee7a18 — `credSources` now returns `(target, migrationSources)`, no ordering coupling.

cmd/auth/auth.go
Badlazzor: Done in 4ee7a18 — see thread above; the `(target, migrationSources)` split removes the ordering contract.

cmd/auth/auth.go
Badlazzor: Done in fa05685 — routed through `xceleratconfig.ConfigFile`, `ccacheconfig.ConfigFile`, `multiplatformconfig.FilePath`, and `paths.GradleInitScriptFile()`.

cmd/auth/auth.go
Badlazzor: Done in fa05685 — absent sources are now hidden by default; re-run `auth list --debug` for the full audit. Errors/partials always print.

cmd/auth/auth.go
Badlazzor: Agreed — keeping 4-char tail. Local ad-hoc + masked is acceptable.

cmd/auth/auth.go
Badlazzor: Done in 4ee7a18 — `auth get` → `auth list`.

cmd/auth/auth.go
Badlazzor: Yes — `zalando/go-keyring` uses libsecret / freedesktop Secret Service on Linux; macOS uses Keychain Services.

cmd/auth/auth.go
```diff
@@ -0,0 +1,514 @@
...
+
+	return "~/" + filepath.ToSlash(rel)
+}
+
+//nolint:gochecknoglobals
+var (
+	authTokenLiteralRE = regexp.MustCompile(`authToken(?:\s*=\s*|\.set\()"[^"]+"`)
+	quotedStringRE     = regexp.MustCompile(`"[^"]+"`)
+
+	reactivateHints = map[string]string{
```
zsolt-marta-bitrise: Please still don't depend on paths here, it's ode duplication to paths.go and a changing path can render the reactivation hint ineffective. You can introduce common IDs for map keys or refer to the constants defined in that package.

zsolt-marta-bitrise (approved):

cmd/gradle/enable_for_gradle.go
```diff
@@ -82,7 +83,15 @@ func EnableForGradleCmdFn(logger log.Logger, gradleHomePath string, envProvider
...
 	activateGradleParams.TestDistro.Enabled = false
 
-	authConfig, _ := configcommon.ReadAuthConfigFromEnvironments(envProvider)
+	if cliPath, exeErr := os.Executable(); exeErr == nil {
+		activateGradleParams.CLIPath = cliPath
+	}
+
+	authConfig, err := configcommon.ResolveAuthConfig(envProvider)
+	if err != nil {
+		return fmt.Errorf(FmtErrorEnableForGradle, fmt.Errorf(gradleconfig.ErrFmtReadAuthConfig, err))
```
zsolt-marta-bitrise, line 92: [bitrise-plugin-ai-reviewer]: cmd/gradle/enable_for_gradle.go:92:0:9cbb0498

[//]: # ([skip ci])


**🐛 Bug**

Error message is double-wrapped. The error is wrapped with ErrFmtReadAuthConfig ("resolve auth config: %w") and then wrapped again with FmtErrorEnableForGradle ("adding Gradle plugins failed: %w"), resulting in redundant "resolve auth config" text in the final error message. Compare with restoreGradleConfigurationCache.go line 50 which uses a single wrapping.

🔄 Suggestion:
```suggestion
		return fmt.Errorf(FmtErrorEnableForGradle, err)
```

zsolt-marta-bitrise (commented):

PR #351, opened 2026-06-03
feat: ACI-5027 productionize activate --interactive wizard

Badlazzor:
> **Stacked on #350 (ACI-5028 keychain).** This branch contains the A2 commits — the diff will look cleaner once A2 merges and I rebase onto main.

## Summary

Finishes the [ACI-4917](https://bitrise.atlassian.net/browse/ACI-4917) PoC into the production wizard required by [ACI-5027](https://bitrise.atlassian.net/browse/ACI-5027).

The PoC commit (`cac1fc1`, Viktor) is cherry-picked as-is so the original author keeps attribution. A second commit on top adds three things the spec required:

- **ccache** as a 4th tool option (was Gradle / Bazel / Xcode only).
- **Persistence to the OS keychain** via the new `internal/auth/keychain` from #350. Downstream activation reads credentials back through `ResolveAuthConfig`.
- **Migration prompt** for users who already have `BITRISE_BUILD_CACHE_AUTH_TOKEN` + `WORKSPACE_ID` in env — wizard imports them to the keychain and tells the user to drop them from shell rc files. (Spec Q4 = c: handle only inside `activate --interactive`.)

Behaviour summary for `resolveCredentials`:

| State going in | Wizard does |
| --- | --- |
| Keychain populated | Reuses silently, no re-prompt |
| Env populated, keychain empty | Imports env → keychain, prints removal hint |
| Nothing set | Prompts (masked token), persists to keychain |
| Keychain save fails (D-Bus down etc) | Non-fatal, continues for the current run only |

## Test plan

- [x] `go build ./...` clean
- [x] `make lint` — 0 issues
- [x] `make test-unit` — green
- [x] New unit tests cover all four credential states + tool-prompt with ccache + existing PoC tests updated for 4-tool index

## Out of scope (referenced for reviewer context)

- Keychain primitives themselves → #350 (ACI-5028)
- Repo-controlled config merge → G2 / ACI-5052 (M3)

[ACI-4917]: https://bitrise.atlassian.net/browse/ACI-4917?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ACI-5027]: https://bitrise.atlassian.net/browse/ACI-5027?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


zsolt-marta-bitrise:
[//]: # ([skip ci])



## Summary
Centralizes auth environment variable names and refactors credential resolution to return the source (keychain, env vars, JWT, etc.). Updates 20+ call sites to handle the new three-value return signature. Adds GetKeychainCredentials() helper and AuthSource enum for tracking credential provenance.


## Walkthrough
| File | Summary |
|------|---------|
| internal/config/common/auth.go, cache_config.go | Added EnvAuthToken, EnvWorkspaceID, EnvJWT constants; added AuthSource enum; modified ResolveAuthConfig() to return (CacheAuthConfig, AuthSource, error); added GetKeychainCredentials() helper |
| cmd/auth/auth.go, cmd/gradle/*.go, cmd/xcode/*.go | Updated ResolveAuthConfig() call sites to handle new three-value return signature; use centralized env var constants |
| internal/config/gradle/*.go, bazel/*.go, xcelerate/*.go, ccache/config.go | Updated ResolveAuthConfig() call sites to discard AuthSource return value |
| pkg/file/helper.go, pkg/reactnative/activator.go | Updated ResolveAuthConfig() call sites to handle new signature |
| cmd/common/activate_interactive*.go, tests | New interactive wizard implementation with keychain credential persistence and migration support |

cmd/common/activate_interactive.go
```diff
@@ -0,0 +1,559 @@
...
+		}
+
+		value = strings.TrimSpace(value)
+		if value != "" {
+			return value, nil
+		}
+
+		fmt.Fprintf(p.out, "%s cannot be empty.\n", label)
+	}
+}
```
zsolt-marta-bitrise: [bitrise-plugin-ai-reviewer]: cmd/common/activate_interactive.go:436:454:f893cca8

[//]: # ([skip ci])


**🐛 Bug**

The function has an infinite loop when stdin is closed (EOF). The readPassword() function returns ("", nil) on EOF, so after printing the "cannot be empty" error, the loop continues and tries to read again, causing an infinite loop. This is inconsistent with promptRequiredLine() which properly checks for EOF after the error message.

🔄 Suggestion:
```suggestion
func promptRequiredSecret(p *prompter, label string) (string, error) {
	for {
		fmt.Fprintf(p.out, "%s: ", label)

		value, err := p.readPassword()
		fmt.Fprintln(p.out)

		if err != nil {
			return "", err
		}

		value = strings.TrimSpace(value)
		if value != "" {
			return value, nil
		}

		fmt.Fprintf(p.out, "%s cannot be empty.\n", label)

		// Check if stdin was closed to prevent infinite loop
		if errors.Is(err, io.EOF) {
			return "", fmt.Errorf("%s not provided (stdin closed)", label)
		}
	}
}
```

zsolt-marta-bitrise (commented):

cmd/common/activate_interactive_huh.go
```diff
@@ -0,0 +1,144 @@
...
+
+	switch source {
+	case credsSourceKeychain:
+		logger.TInfof("Using credentials from the OS keychain.")
+	case credsSourceEnv:
+		logger.TInfof("Imported BITRISE_BUILD_CACHE_AUTH_TOKEN + WORKSPACE_ID from env into the OS keychain.")
+		logger.Infof("You can now remove them from your shell rc files.")
+		_ = persistCredentials(kc, workspaceID, authToken)
+	case credsSourceNone:
+		_ = persistCredentials(kc, workspaceID, authToken)
```
zsolt-marta-bitrise: [bitrise-plugin-ai-reviewer]: cmd/common/activate_interactive_huh.go:97:0:a50870a6

[//]: # ([skip ci])


**🐛 Bug**

The success message is printed even if keychain save fails. The error from persistCredentials is silently ignored, so the user sees "Credentials saved" even when the save actually failed (e.g., D-Bus down). This is misleading. The plain wizard correctly checks the error and prints different messages based on success/failure.

<details>

```
In cmd/common/activate_interactive_huh.go at line 97, Check the error from persistCredentials and only print the success message if the save actually succeeded. Print an error message if the save fails, similar to the plain wizard's resolveCredentials function.
```

</details>

cmd/common/activate_interactive_huh.go
```diff
@@ -0,0 +1,144 @@
...
+	)
+
+	if err := huh.NewForm(groups...).Run(); err != nil {
+		return fmt.Errorf("interactive wizard: %w", err)
+	}
+
+	switch source {
+	case credsSourceKeychain:
+		logger.TInfof("Using credentials from the OS keychain.")
+	case credsSourceEnv:
```
zsolt-marta-bitrise: [bitrise-plugin-ai-reviewer]: cmd/common/activate_interactive_huh.go:92:0:a50870a6

[//]: # ([skip ci])


**🐛 Bug**

The message "Imported ... into the OS keychain" is printed even if the keychain save fails. The error from persistCredentials is silently ignored, so the user is told they can remove env vars even when the save actually failed. This is misleading. The plain wizard correctly checks the error and only prints the removal hint if the save succeeds.

<details>

```
In cmd/common/activate_interactive_huh.go at line 92, Check the error from persistCredentials. Only print the "remove from shell rc files" message if the save actually succeeded. If the save fails, print an error message instead.
```

</details>

zsolt-marta-bitrise (commented):

cmd/common/activate_interactive.go
```diff
@@ -0,0 +1,198 @@
...
+	"github.com/bitrise-io/bitrise-build-cache-cli/v2/internal/config/xcelerate"
+	"github.com/bitrise-io/bitrise-build-cache-cli/v2/internal/utils"
+	ccachepkg "github.com/bitrise-io/bitrise-build-cache-cli/v2/pkg/ccache"
+)
+
+//nolint:gochecknoglobals
+var interactiveFlag bool
+
+const (
+	envWorkspaceID = "BITRISE_BUILD_CACHE_WORKSPACE_ID"
```
zsolt-marta-bitrise: Can we use auth package for these? I want to avoid having these envs defined in many places + we have the JWT env. See other thread for full suggestion

cmd/common/activate_interactive.go
```diff
@@ -0,0 +1,198 @@
...
+type wizard interface {
+	Run(ctx context.Context) error
+}
+
+func selectWizard() wizard {
+	if term.IsTerminal(int(os.Stdin.Fd())) {
+		return &huhWizard{}
+	}
+
+	return &plainWizard{prompter: newDefaultPrompter()}
```
zsolt-marta-bitrise: Is this still realistic? Do we really have to support this use case? The code for that is quite large

cmd/common/activate_interactive_plain.go
```diff
@@ -0,0 +1,247 @@
...
+
+func resolveCredentials(p *prompter, kc keychainStore) (string, string, error) {
+	creds, err := kc.Load()
+	if err == nil && creds.AuthToken != "" && creds.WorkspaceID != "" {
+		fmt.Fprintln(p.out, "Reusing credentials already stored in the OS keychain.")
+
+		return creds.WorkspaceID, creds.AuthToken, nil
+	}
+
+	envToken := os.Getenv(envAuthToken)
```
zsolt-marta-bitrise: Please use the config common package that resolves credentials. Actually, now we have the keychain referenced both here and in the `auth list` command, can we move the keychain resolution also down one level into the config package's ResolveAuthConfig (and return the source found)? That way we can have the auth stuff in one place and not readd the keychain stuff into these package too

cmd/common/activate_interactive_plain.go
```diff
@@ -0,0 +1,247 @@
...
+	envWS := os.Getenv(envWorkspaceID)
+
+	if envToken != "" && envWS != "" {
+		fmt.Fprintln(p.out)
+		fmt.Fprintln(p.out, "Found BITRISE_BUILD_CACHE_AUTH_TOKEN + BITRISE_BUILD_CACHE_WORKSPACE_ID in env.")
+		fmt.Fprintln(p.out, "Importing them into the OS keychain so you can remove them from your shell rc files.")
```
zsolt-marta-bitrise: [bitrise-plugin-ai-reviewer]: cmd/common/activate_interactive_plain.go:107:0:a8f13c3f

[//]: # ([skip ci])


**🐛 Bug**

This message is printed before attempting to save credentials to the keychain. If the save fails (line 109), the user is told they can remove the env vars, but they cannot because the save failed and the credentials are still needed. The message should only be printed after a successful save, or should be changed to indicate an attempt rather than a completed action.

🔄 Suggestion:
```suggestion
		if err := kc.Save(keychain.Credentials{AuthToken: envToken, WorkspaceID: envWS}); err != nil {
			fmt.Fprintf(p.out, "Could not save to keychain (%v). Continuing with env values for this run only.\n", err)

			return envWS, envToken, nil
		}

		fmt.Fprintln(p.out, "✅ Credentials saved to the OS keychain.")
		fmt.Fprintln(p.out, "Imported BITRISE_BUILD_CACHE_AUTH_TOKEN + BITRISE_BUILD_CACHE_WORKSPACE_ID from env into the OS keychain.")
		fmt.Fprintln(p.out, "You can now remove them from your shell rc files.")
```

zsolt-marta-bitrise (commented):

cmd/common/activate_interactive.go
```diff
@@ -0,0 +1,198 @@
...
+	return nil
+}
+
+func runInteractiveGradle(logger log.Logger, envs map[string]string, pushEnabled bool) error {
+	gradleHome, err := pathutil.NewPathModifier().AbsPath("~/.gradle")
+	if err != nil {
+		return fmt.Errorf("expand Gradle home path: %w", err)
+	}
+
+	params := gradleconfig.DefaultActivateGradleParams()
```
zsolt-marta-bitrise, line 89: [bitrise-plugin-ai-reviewer]: cmd/common/activate_interactive.go:101:0:f893cca8

[//]: # ([skip ci])


**🐛 Bug**

The `CLIPath` field is not being set in the interactive wizard, unlike in the regular `activate_gradle.go` command. This means the template will use the default value "bitrise-build-cache" instead of the actual executable path. While there's a fallback, this could fail if the CLI is not in the PATH or installed in a non-standard location.

🔄 Suggestion:
```suggestion
	params := gradleconfig.DefaultActivateGradleParams()
	params.Cache.PushEnabled = pushEnabled

	if cliPath, exeErr := os.Executable(); exeErr == nil {
		params.CLIPath = cliPath
	}

	if err := gradleconfig.Activate(
```

zsolt-marta-bitrise (commented):

cmd/common/activate_interactive.go
Badlazzor: Not really, inherited from viktor's original POC code, thought it might still be relevant. But you are right, not really. (It wasn't claude trying to sneak in more code ;) )

cmd/common/activate_interactive.go
Badlazzor: Done in 6499ba9 — added `configcommon.EnvAuthToken` / `EnvWorkspaceID` / `EnvJWT` and replaced all programmatic lookups (interactive wizard, `auth list` probes, `cache_config` redactor). User-facing copy that mentions the names by hand stays for readability.

cmd/common/activate_interactive_plain.go
Badlazzor: `resolveCredentials` (and the whole plain wizard) is gone in f3f777a. `ResolveAuthConfig` in configcommon already does env→keychain→multiplatform—the duplication is fixed by removing the parallel wizard path. Returning the resolved source from `ResolveAuthConfig` is a separate refactor I'll pick up if/when another caller needs it (currently only the wizard's `loadStartingCredentials` does keychain-first semantics that differ from ResolveAuthConfig's env-first ordering).

zsolt-marta-bitrise: Why do we need different paths for credential resolutions? Can we not always check for keychain first?

Badlazzor: ~We can if we force all users to use that~

Badlazzor: I would argue that env vars should still come first, as that is an industry standard. We need the 2 exported functions for different scenarios:
- ResolveAuthConfig: this is the preferred one, should be used unless...
- ResolveStoredCredentials: For "what creds already live somewhere stable?" semantics (wizard's reuse-or-import logic).

cmd/common/activate_interactive.go
```diff
@@ -0,0 +1,190 @@
...
```
zsolt-marta-bitrise, line 89: [bitrise-plugin-ai-reviewer]: cmd/common/activate_interactive.go:89:0:8a4f55d9

[//]: # ([skip ci])


**🐛 Bug**

Cache is not being enabled for the selected tool. The function logs "✅ Bitrise Build Cache activated for Gradle" but params.Cache.Enabled defaults to false, so the cache plugin won't actually be activated. This should be set to true since the user explicitly selected Gradle in the wizard.

🔄 Suggestion:
```suggestion
	params := gradleconfig.DefaultActivateGradleParams()
	params.Cache.Enabled = true
	params.Cache.PushEnabled = pushEnabled
```

zsolt-marta-bitrise (commented):

zsolt-marta-bitrise (approved):
LGTM, but please check https://github.com/bitrise-io/bitrise-build-cache-cli/pull/351#discussion_r3461159770

PR #354, opened 2026-06-08
[NOJIRA] feat: report RN child invocations inline; drop per-child relation calls

imrekel:
## What

The React Native wrapper now reports its child invocations (gradle/xcode/ccache sub-builds) **inline** on its own invocation, instead of each child PUTing a separate parent→child relation to the analytics backend.

## Why

The backend's PubSub messages for RN invocations were missing the `children` field. Children were only ever reported out-of-band via `PUT /v1/invocations/{parent}/children/{child}`, which the publish path never read. Reporting children inline on the parent (combined with backend hydration) closes that gap and removes redundant per-child network calls.

## How

The child IDs + build tools are **already persisted** in the parent's local childstats ledger (`~/.bitrise/cache/invocations/<parent>/`), written by each child as it finishes. They were read for hit-rate aggregation but the IDs were discarded.

- `childstats.Summary` now also surfaces the full lineage (`Children []ChildRef`), populated for **every** well-formed entry — including no-activity children, so the inline list matches what the relation calls covered.
- The RN post-run hook maps it onto the wrapper invocation's new `Children []InvocationRef` field (`internal/analytics/multiplatform`).

With lineage reported by the parent, the redundant child-driven relation calls are removed:

- **xcode** — dropped `sendRelation`/`relationSender` wiring; `saveInvocationAndRelation` → `saveInvocation`. The child still writes its ledger entry, which the parent enumerates.
- **ccache** — dropped `registerInvocationRelation` and the now-unused `registry`.

### Intentionally kept: the gradle path

`register-child-invocation` (invoked by the external gradle-plugins repo) is left in place — removing it would break gradle builds. Gradle children are still covered by the parent's inline list via the same ledger, and the backend upsert is idempotent on `invocation_id`, so the temporary double-report is harmless. Fully dedup'ing gradle is a follow-up in the gradle-plugins repo.

## Deploy order

Requires the backend to hydrate children before publishing — **bitrise-io/multiplatform-analytics-service#21**. Deploy that first: once the CLI stops sending xcode/ccache relations, the inline list is their only source, and the backend needs hydration to publish it.

## Tests

- New `cmd/xcode/xcodebuild_savechild_test.go` covers `saveInvocation`'s ledger-write behavior (replaces the removed relation tests).
- New childstats test asserts lineage includes no-activity children and excludes malformed/aggregate-skipped entries.
- Build, gofmt, and `golangci-lint` are clean on the affected packages. (`cmd/xcode/Test_xcodebuildCmdFn` is a pre-existing network-dependent test, unaffected by this change.)


zsolt-marta-bitrise:
[//]: # ([skip ci])



## Summary
React Native wrapper now reports child invocations (gradle/xcode/ccache sub-builds) inline on the parent invocation instead of via separate relation API calls. Children are enumerated from the parent's local childstats ledger and included in the analytics payload, eliminating redundant per-child network calls.


## Walkthrough
| File | Summary |
|------|---------|
| internal/analytics/multiplatform/invocations.go | Added Children field to Invocation and InvocationRunStats; created InvocationRef type for child references |
| pkg/common/childstats/childstats.go | Added ChildRef struct and Children field to Summary; Compute() now populates lineage for all well-formed entries including no-activity children |
| pkg/reactnative/post_run_deps.go | Extract child references from childstats summary and map to InvocationRef; pass children list inline to invocation creation |
| cmd/xcode/xcodebuild.go | Removed relationSender interface and sendRelation() calls; renamed saveInvocationAndRelation() to saveInvocation(); write to childstats ledger instead |
| pkg/ccache/storage_helper.go | Removed InvocationRegistry field and registerInvocationRelation() method; added writeChildStatsLedger() for ledger-based reporting |
| scripts/react_e2e_assert_analytics.sh, cmd/xcode/*_test.go, pkg/common/childstats/childstats_test.go | Updated tests: removed relation API assertions; added ledger-write and lineage tests; updated analytics error message check |

Badlazzor (approved):

PR #352, opened 2026-06-03
feat: ACI-5034 add bitrise-build-cache health subcommand

Badlazzor:
> **Stacked on #350 (ACI-5028 keychain).** Branch sits on top of A2 so the auth check can use the new keychain package directly. Diff will compress once A2 merges.

## Summary

[ACI-5034](https://bitrise.atlassian.net/browse/ACI-5034) (C1). New top-level \`bitrise-build-cache health\` command — one-shot "is everything working?" view of the local setup. Solves P5 from the M1 plan.

\`\`\`
$ bitrise-build-cache health
Bitrise Build Cache CLI v2.8.3

  ✓ xcelerate-proxy    running (pid 87583)
  ! ccache-helper      not running (no socket file). Run \`bitrise-build-cache ccache start-storage-helper\` if you're using ccache.
  ✓ auth               OS keychain, workspace=322a005426441b60
  ✓ cli-version        current=v2.8.3 (up to date)

Overall: ok
\`\`\`

\`--json\` emits the same data machine-readable for tooling (lays groundwork for C3 self-check step + IDE pickup).

## Naming

The spec said \`status\` but \`bitrise-build-cache status\` already exists with different semantics (per-feature activation query used by step integrations). Renamed this one to \`health\` to keep both. Open to renaming back if you'd rather merge the two — say the word and I'll fold them.

## Checks today

| Check | Detection |
| --- | --- |
| xcelerate-proxy | \`~/.bitrise-xcelerate/proxy.pid\` exists + \`syscall.Kill(pid, 0)\` succeeds |
| ccache-helper | \`BITRISE_CCACHE_IPC_SOCKET_PATH\` or default \`\$TMPDIR/ccache-ipc.sock\` exists + accepts a unix dial |
| auth | keychain.Load → env-var fallback → missing |
| cli-version | github.com/bitrise-io/bitrise-build-cache-cli releases/latest, 3 s timeout. Local builds (\`devel\`/\`+dirty\`) skip the network call. |

Unreachable network → warn (not error). Missing creds / dead proxy pid → error. No-pid / no-socket → warn (the user might simply not be using that tool).

## Test plan

- [x] \`make lint\` 0 issues
- [x] \`make test-unit\` green
- [x] Unit tests cover Overall() severity, local-build detection, all xcelerate-proxy paths, auth keychain → env → missing fallthrough, version compare incl. dirty handling, ccache no-socket warn
- [x] Smoke test on macOS: defaults to colored output, \`--json\` emits valid JSON

[...]

PR #355, opened 2026-06-10
feat: ACI-5034 + ACI-5035 unified bitrise-build-cache doctor command

Badlazzor:
> **Stacked on #350 (ACI-5028 keychain).** Base = \`aci-5028-keychain-auth\`. Diff = doctor + tests only.

## Summary

Single \`bitrise-build-cache doctor\` command — supersedes the separate \`health\` (originally [ACI-5034](https://bitrise.atlassian.net/browse/ACI-5034) / #352) and \`doctor\` ([ACI-5035](https://bitrise.atlassian.net/browse/ACI-5035)) commands.

Peer feedback was the right call: the two surfaces had 3 of 8 checks duplicated and the "fast vs heavy" split was speculative — most checks are sub-millisecond. Single command, single user mental model, every check runs every invocation.

\`\`\`
$ bitrise-build-cache doctor
Bitrise Build Cache - doctor
CLI version: v2.8.3

  ✓ auth                   OS keychain, workspace=322a005426441b60
  ✓ keychain-smoke         Set/Get/Delete round-trip OK
  ! xcelerate-proxy        stale pid file: pid 74360 not running — fixable
      ↳ rerun with --fix to repair
  ! ccache-helper          not running (no socket file). Run \`bitrise-build-cache ccache start-storage-helper\` if you build C/C++.
  ! ccache-binary          ccache binary not found in PATH. Install via \`brew install ccache\` if you build C/C++.
  ✓ log-dirs               all log dirs present + writable
  ✓ xcelerate-xcconfig     xcelerate dir exists but no xcconfig files (no Xcode local activation yet)
  ✓ cli-version            current=v2.8.3 (up to date)

Overall: warn
\`\`\`

## Checks (8 in one place)

| Check | What | --fix repair |
| --- | --- | --- |
| auth | keychain → env fallback → missing | — (use \`auth set\` or wizard) |
| keychain-smoke | Set/Get/Delete round-trip via doctor-only service/account | — |
| xcelerate-proxy | pid file alive (covers health's "is running" + doctor's "stale pid") | rm stale pid file |
| ccache-helper | socket probe via DialContext | rm orphan socket |
| ccache-binary | which ccache | — (user installs) |
| log-dirs | ~/.local/state/{xcelerate,ccache}/logs writable | mkdir missing |
| xcelerate-xcconfig | file sanity | — |
| cli-version | current vs latest GitHub release | — (use \`update\` or \`brew upgrade\`) |

## Flags

- \`--fix\` — apply safe repairs.
- \`--json\` — machine-readable output.
- \`--no-update-check\` — skip the GitHub release lookup (the only network call, ~3s timeout).

Exit code is non-zero when overall state is error.

## Heads-up — pre-existing test flake on dev machines

A few \`cmd/{bazel,gradle,xcode}\` tests fail locally on dev machines that have keychain credentials populated. Root cause is in A2 (#350): \`ResolveAuthConfig\` consults keychain first, which makes some auth-sensitive tests behave differently than on CI (where keychain access returns ErrNotFound). CI is green. Worth tracking with a test-only bypass env (e.g. \`BITRISE_BUILD_CACHE_SKIP_KEYCHAIN=1\`) in the A2 PR.

## Test plan

- [x] \`make lint\` — 0 issues
- [x] \`make test-unit\` on the new package — green; 23 sub-tests cover Overall(), local-build detection, all 8 checks (incl. happy + sad paths), Run() with ApplyFixes + SkipUpdateCheck
- [x] Smoke test on macOS: doctor / doctor --fix / doctor --json / doctor --no-update-check

[ACI-5034]: https://bitrise.atlassian.net/browse/ACI-5034?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ACI-5035]: https://bitrise.atlassian.net/browse/ACI-5035?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


zsolt-marta-bitrise:
[//]: # ([skip ci])



## Summary
Unified `bitrise-build-cache doctor` command with 8 health checks (auth, keychain-smoke, xcelerate-proxy, ccache-helper, ccache-binary, log-dirs, xcelerate-xcconfig, cli-version). Supports --fix for safe repairs, --json for machine output, and --no-update-check to skip GitHub release lookup. Exit code is non-zero on errors.


## Walkthrough
| File | Summary |
|------|---------|
| cmd/doctor/doctor.go | New doctor command with human-readable + JSON output, --fix repairs, proper exit codes |
| internal/doctor/{check_*.go,doctor.go} | 8 health checks: auth, keychain-smoke, xcelerate-proxy, ccache-helper, ccache-binary, log-dirs, xcelerate-xcconfig, cli-version with Report aggregation + Run() method |
| internal/config/{ccache,xcelerate}/config.go, xcelerate/paths.go | Centralized socket path resolution via ResolveIPCSocketPath() + ResolveProxySocketPath() helpers |
| internal/auth/keychain/keychain.go | Exported NewBackend() for doctor keychain smoke test |
| internal/config/common/auth.go | Exported AuthLoader interface + GetKeychainCredentialsWith() for doctor auth check |
| internal/paths/paths.go | Added ProxyPidFile(), CcacheSocketPath(), XcelerateLogDir(), CcacheLogDir() methods + constants |
| main.go | Registered doctor command via blank import |
| internal/doctor/doctor_test.go, internal/config/common/auth_test.go, CLAUDE.md | Tests + documentation updates |

cmd/doctor/doctor.go
```diff
@@ -0,0 +1,157 @@
...
+	SilenceUsage: true,
+	RunE: func(cmd *cobra.Command, _ []string) error {
+		out := cmd.OutOrStdout()
+
+		report := doctorpkg.NewRunner().Run(cmd.Context(), doctorpkg.Options{
+			ApplyFixes:      fixFlag,
+			SkipUpdateCheck: skipUpdateCheckFlag,
+		})
+
+		if jsonOutput {
```
zsolt-marta-bitrise, line 45: [bitrise-plugin-ai-reviewer]: cmd/doctor/doctor.go:45:0:13fa1ad5

[//]: # ([skip ci])


**🐛 Bug**

When --json flag is used, the function returns immediately after writeJSON() without checking if report.Overall() is StateError. This means the exit code will be 0 even when there are errors in the report, violating the requirement that "Exit code is non-zero when overall state is error".

🔄 Suggestion:
```suggestion
		if jsonOutput {
			if err := writeJSON(out, report); err != nil {
				return err
			}
			if report.Overall() == doctorpkg.StateError {
				return errors.New("doctor reported errors")
			}
			return nil
		}

		writeHuman(out, report, fixFlag, colorEnabled(out))

		if report.Overall() == doctorpkg.StateError {
			return errors.New("doctor reported errors")
		}

		return nil
```

zsolt-marta-bitrise (commented):

cmd/doctor/doctor.go
zsolt-marta-bitrise, line 45: [bitrise-plugin-ai-reviewer]: cmd/doctor/doctor.go:45:0:cbe4a6f9

[//]: # ([skip ci])


**🐛 Bug**

Exit code bug: When --json flag is used, the function returns without checking report.Overall() for errors. This causes the exit code to be 0 even when there are errors in the report. The error check at line 51 is only executed for non-JSON output.

🔄 Suggestion:
```suggestion
		if jsonOutput {
			if err := writeJSON(out, report); err != nil {
				return err
			}

			if report.Overall() == doctorpkg.StateError {
				return errors.New("doctor reported errors")
			}

			return nil
		}
```

zsolt-marta-bitrise (commented):

internal/doctor/check_ccache.go
```diff
@@ -0,0 +1,77 @@
...
+	"os"
+	"path/filepath"
+	"time"
+)
+
+func (d *Doctor) ccacheHelperCheck() Check {
+	return Check{
+		Name: "ccache-helper",
+		Diagnose: func(ctx context.Context) Result {
+			socketPath := d.Envs["BITRISE_CCACHE_IPC_SOCKET_PATH"]
```
zsolt-marta-bitrise: Do we have a `paths` entry for this? How can we ensure that the doctor's way of searching for the socket is the same that the ccache package and the storage helper is using?

cmd/doctor/doctor.go
```diff
@@ -0,0 +1,159 @@
...
+	for _, it := range r.Items {
+		fmt.Fprintf(w, "  %s %-22s %s\n", c.icon(it.Result.State), it.Name, it.Result.Detail)
+
+		switch {
+		case it.FixResult != nil:
+			fmt.Fprintf(w, "      %s↳ fixed:%s %s\n", c.green, c.reset, *it.FixResult)
+		case it.FixError != "":
+			fmt.Fprintf(w, "      %s↳ fix failed:%s %s\n", c.red, c.reset, it.FixError)
+		case !fixed && it.Result.Fixable:
+			fmt.Fprintf(w, "      %s↳%s rerun with --fix to repair\n", c.yellow, c.reset)
```
zsolt-marta-bitrise, line 103: I like this approach that we print a comprehensive report, with formatting, and a separate `--fix` command

internal/doctor/check_cli_version.go
```diff
@@ -0,0 +1,88 @@
...
+	return Check{
+		Name: "cli-version",
+		Diagnose: func(ctx context.Context) Result {
+			current := d.CLIVersion
+			if current == "" {
+				current = "devel"
+			}
+
+			if isLocalBuild(current) {
+				return Result{State: StateOK, Detail: "current=" + current + " (local build)"}
```
zsolt-marta-bitrise: I would warn about this, customers should never build their own CLI unless for a hotfix. In that case we should prompt them to update

internal/doctor/check_cli_version.go
```diff
@@ -0,0 +1,88 @@
...
+				return Result{State: StateOK, Detail: "current=" + current}
+			}
+
+			if current == latest || strings.TrimPrefix(current, "v") == strings.TrimPrefix(latest, "v") {
+				return Result{State: StateOK, Detail: "current=" + current + " (up to date)"}
+			}
+
+			return Result{
+				State:  StateWarn,
+				Detail: fmt.Sprintf("current=%s, latest=%s — run `bitrise-build-cache update` or `brew upgrade`", current, latest),
```
zsolt-marta-bitrise: Does the update command work if they have installed via homebrew + vice-versa, does the brew upgrade command work if they have installed in a different way?

internal/doctor/check_keychain.go
```diff
@@ -0,0 +1,54 @@
...
+			}
+
+			got, err := d.Keyring.Get(smokeServiceName, smokeAccountName)
+			if err != nil || got != secret {
+				_ = d.Keyring.Delete(smokeServiceName, smokeAccountName)
+				if err != nil {
+					return Result{State: StateError, Detail: "keychain Get failed: " + err.Error()}
+				}
+
+				return Result{State: StateError, Detail: "keychain Get returned mismatched value (stale entry from a previous run with a failed Delete?)"}
```
zsolt-marta-bitrise, line 44: Is this user actionable? Wouldn't this surface when they set the auth, and when they call `auth token` ? What's the purpose of checking a _different_ key?

internal/doctor/check_ccache.go
```diff
@@ -0,0 +1,77 @@
...
+	"fmt"
+	"io/fs"
+	"net"
+	"os"
+	"path/filepath"
+	"time"
+)
+
+func (d *Doctor) ccacheHelperCheck() Check {
+	return Check{
```
zsolt-marta-bitrise, line 20: I like the format of defining these Checks in a self-contained way

internal/doctor/doctor.go
```diff
@@ -0,0 +1,139 @@
...
+	return &Doctor{
+		OsProxy:            osProxy,
+		Envs:               utils.AllEnvs(),
+		CLIVersion:         common.GetCLIVersion(nil),
+		HTTPClient:         &http.Client{Timeout: 3 * time.Second},
+		AuthLoader:         keychain.New(),
+		Keyring:            keychain.NewBackend(),
+		XcelerateProxyDir:  func() string { return xcelerate.DirPath(osProxy) },
+		LookPath:           exec.LookPath,
+		StateDirCandidates: []string{"~/.local/state/xcelerate/logs", "~/.local/state/ccache/logs"},
```
zsolt-marta-bitrise: Again, please use the paths package. Please also check these before asking for a review, I had to point these out multiple times now across many PRs. 

Code reuse and separation of concerns is important and we have a ton of features and files in the CLI now, it's imperative we can keep an eye on where and what we generate on users' machines, especially important now with the local dx.

internal/doctor/check_logdirs.go
```diff
@@ -0,0 +1,139 @@
...
+
+			return s
+		}
+		if out.Missing != "" {
+			s.Missing = append(s.Missing, out.Missing)
+		}
+		if out.NotWritable != "" {
+			s.NotWritable = append(s.NotWritable, out.NotWritable)
+		}
+		if out.WrongOwner != "" {
```
zsolt-marta-bitrise, line 72: Yeah, good idea to check we can write the logs before something happens that would've needed a log

internal/doctor/check_xcelerate.go
```diff
@@ -0,0 +1,57 @@
...
+
+			if err := syscall.Kill(pid, 0); err != nil {
+				return Result{
+					State:   StateWarn,
+					Detail:  fmt.Sprintf("stale pid file: pid %d not running — fixable", pid),
+					Fixable: true,
+				}
+			}
+
+			return Result{State: StateOK, Detail: fmt.Sprintf("running (pid %d)", pid)}
```
zsolt-marta-bitrise: Let's also check the socket so we can connect to it (similar to ccache proxy)

internal/doctor/check_ccache.go
Badlazzor: Done in 5065b92 — added `ccacheconfig.EnvIPCSocketPath` + `ResolveIPCSocketPath(override, envs, osProxy)`. Activator and doctor now share the same chain (override → env → tmp/ccache-ipc.sock), so the doctor can't drift from where the storage helper actually binds.

internal/doctor/check_cli_version.go
Badlazzor: Done in 5065b92 — local builds (`devel` / `+commit` / `+dirty`) now resolve to `StateWarn`, with the detail pointing at the installer script + brew formula. Saving `Error` for genuine breakage since hotfix builds are a legitimate (if rare) scenario.

internal/doctor/check_cli_version.go
Badlazzor: Yes to both — `bitrise-build-cache update` in `cmd/update/update.go` already detects the install method via `internal/updater/{brew,manual}.go` and dispatches to `brew update && brew upgrade` or re-runs installer.sh against the same bindir. Hint in 5065b92 dropped the "or `brew upgrade`" fork and now just points at `update`.

internal/doctor/check_keychain.go
Badlazzor: User-actionable in the sense that `auth set` would surface the same backend failure — but only when the user reaches for it. doctor's purpose is to flag a broken keychain backend (D-Bus down, locked login keychain) BEFORE `auth set` / `auth token` hits it with a less obvious "save credentials failed" error. The smoke key uses a separate service+account namespace (`bitrise-build-cache-doctor` / `smoketest`) so the round-trip can't stomp on the user's real creds. Happy to drop it if you'd rather defer to the failure-at-use-time approach — but I think the preemptive diagnostic is worth the ~30 lines.

internal/doctor/doctor.go
Badlazzor: Done in 5065b92 — `StateDirCandidates` resolves through `paths.FromHome(home).XcelerateLogDir() / .CcacheLogDir()`, ccache + xcelerate socket paths come from per-package `Resolve*Path` helpers that route through `paths`, and `internal/paths` gained the missing filename constants (`ProxyPidFileName`, `CcacheSocketName`, log-dir relatives). Also saved an internal memory rule so future PRs default to this layout without you having to flag it again.

internal/doctor/check_xcelerate.go
Badlazzor: Done in 5065b92 — `check_xcelerate.go` now dials the proxy unix socket after the pid-alive check (500ms timeout, same shape as ccache). Live-pid + non-accepting socket lands as Warn with detail "pid X alive but socket Y not accepting connections".

internal/doctor/doctor.go
```diff
@@ -0,0 +1,149 @@
...
+		d.ccacheHelperCheck(),
+		d.ccacheBinaryCheck(),
+		d.logDirsCheck(),
+	}
+
+	if !skipUpdateCheck {
+		checks = append(checks, d.cliVersionCheck())
+	}
+
+	return checks
```
zsolt-marta-bitrise, line 171: [bitrise-plugin-ai-reviewer]: internal/doctor/doctor.go:134:148:e38334d1

[//]: # ([skip ci])


**🐛 Bug**

Missing xcelerate-xcconfig check. The original code (commit cbe4a6f) includes an xcelerateXcconfigCheck() that validates xcconfig files in the xcelerate directory. This check is not present in the refactored code, reducing the number of checks from 8 to 7 (or 6 without the version check).

<details>

```
In internal/doctor/doctor.go between lines 134 and 148, Add the missing xcelerateXcconfigCheck() to the checks list. Create a new file check_xcelerate_xcconfig.go with the xcelerateXcconfigCheck implementation that validates xcconfig files in the xcelerate directory.
```

</details>

zsolt-marta-bitrise (commented):

internal/config/ccache/config.go
```diff
@@ -90,6 +90,23 @@ func ConfigFile(osProxy utils.OsProxy) string {
...
+func ResolveIPCSocketPath(override string, envs map[string]string, osProxy utils.OsProxy) string {
+	if override != "" {
+		return override
+	}
+	if env := envs[EnvIPCSocketPath]; env != "" {
+		return env
+	}
+
+	return paths.FromHome("").CcacheSocketPath(osProxy.TempDir())
+}
```
zsolt-marta-bitrise, line 108: [bitrise-plugin-ai-reviewer]: internal/config/ccache/config.go:96:108:c0d6ac63

[//]: # ([skip ci])


**📋 Guideline**

The ResolveIPCSocketPath function was added but is not used in NewConfig, which still duplicates the same resolution logic inline. This violates the CLAUDE.md guideline that Resolve...Path helpers must be the single source of resolution for both writers (activator) and inspectors (doctor) to prevent drift.

🔄 Suggestion:
```suggestion
// ResolveIPCSocketPath returns the storage helper's IPC socket path in the same
// order the activator uses: an explicit override → BITRISE_CCACHE_IPC_SOCKET_PATH
// env var → <temp-dir>/ccache-ipc.sock.
func ResolveIPCSocketPath(override string, envs map[string]string, osProxy utils.OsProxy) string {
	if override != "" {
		return override
	}
	if env := envs[EnvIPCSocketPath]; env != "" {
		return env
	}

	return paths.FromHome("").CcacheSocketPath(osProxy.TempDir())
}

func DefaultParams() Params {
	return Params{
		PushEnabled: true,
	}
}

func NewConfig(envs map[string]string, osProxy utils.OsProxy, params Params) (Config, error) {
	authConfig, _, err := common.ResolveAuthConfig(envs)
	if err != nil {
		return Config{}, fmt.Errorf(ErrNoAuthConfig, err)
	}

	ipcEndpoint := ResolveIPCSocketPath(params.IPCSocketPathOverride, envs, osProxy)
```

zsolt-marta-bitrise (commented):

internal/config/xcelerate/paths.go
```diff
@@ -40,3 +40,29 @@ func PathFor(osProxy utils.OsProxy, subpath string) string {
...
 	return PathFor(osProxy, xcelerateConfigFileName)
 }
+
+// EnvProxySocketPath overrides the default xcelerate proxy socket location when set.
+const EnvProxySocketPath = "BITRISE_XCELERATE_PROXY_SOCKET_PATH"
+
+// ResolveProxySocketPath returns the proxy unix socket path in the same order
+// activate uses: explicit override → BITRISE_XCELERATE_PROXY_SOCKET_PATH env var
+// → <temp-dir>/xcelerate-proxy.sock.
+func ResolveProxySocketPath(override string, envs map[string]string, osProxy utils.OsProxy) string {
```
zsolt-marta-bitrise, line 50: [bitrise-plugin-ai-reviewer]: internal/config/xcelerate/paths.go:50:0:c0d6ac63

[//]: # ([skip ci])


**📋 Guideline**

The ResolveProxySocketPath function is defined but not used in NewConfig (internal/config/xcelerate/config.go lines 174-183), which still duplicates the resolution logic manually. This violates the CLAUDE.md rule that all path resolution should go through the paths package to avoid drift between activate and inspector code (doctor, audits).

<details>

```
In internal/config/xcelerate/paths.go at line 50, Update NewConfig in config.go to use ResolveProxySocketPath instead of manually duplicating the resolution logic. This ensures activate and doctor use the same path resolution.
```

</details>

internal/doctor/doctor.go
```diff
@@ -0,0 +1,146 @@
...
+
+func (d *Doctor) checks(skipUpdateCheck bool) []Check {
+	checks := []Check{
+		d.authCheck(),
+		d.keychainSmokeCheck(),
+		d.xcelerateProxyCheck(),
+		d.xcelerateXcconfigCheck(),
+		d.ccacheHelperCheck(),
+		d.ccacheBinaryCheck(),
+		d.logDirsCheck(),
```
zsolt-marta-bitrise: [bitrise-plugin-ai-reviewer]: internal/doctor/doctor.go:135:138:36b1104b

[//]: # ([skip ci])


**🐛 Bug**

The xcelerate-xcconfig check is in the wrong position. According to the PR description output, it should come after log-dirs, not before ccache-helper. The expected order is: auth, keychain-smoke, xcelerate-proxy, ccache-helper, ccache-binary, log-dirs, xcelerate-xcconfig, cli-version.

🔄 Suggestion:
```suggestion
		d.ccacheHelperCheck(),
		d.ccacheBinaryCheck(),
		d.logDirsCheck(),
		d.xcelerateXcconfigCheck(),
```

zsolt-marta-bitrise (commented):

zsolt-marta-bitrise (approved):
I have this suggestion but otherwise LGTM https://github.com/bitrise-io/bitrise-build-cache-cli/commit/5065b923529ff727970180135e8e5cab09cc5a2c#r190104735

PR #362, opened 2026-06-11
feat: ACI-5039 per-tool config-version sidecars + major-bump refresh nudge

Badlazzor:
## Summary

Phase 1 of D3: detect when a tool's on-disk activate config is from a CLI version with a stale config schema, and nudge the user to re-run the matching `bitrise-build-cache activate <tool>`. The nudge fires only on **major** ConfigVersion bumps — minor / patch bumps stay quiet.

## Architecture

- `internal/toolconfig/base.go` — `Tool` enum (gradle / bazel / xcelerate / ccache) + per-tool `ConfigVersion` constants ("1.0.0" baseline) + `Sample` shape the scanner returns.
- `internal/refresh/scanner.go` — walks the four on-disk config locations and returns a `Sample` per tool:
  - `~/.bitrise/cache/gradle/config.json` (new sidecar — init.d script is not JSON, sidecar holds the version)
  - `~/.bitrise/cache/bazel/config.json` (new sidecar — ~/.bazelrc is co-owned with user config, sidecar holds the version)
  - `~/.bitrise-xcelerate/config.json` (existing config; gained `ConfigVersion` + `WrittenAt`)
  - `~/.bitrise/cache/ccache/config.json` (existing config; same new fields)
- `internal/refresh/notify.go` — `needsNudge(stored, current)` returns true when `semver.Major(stored) < semver.Major(current)`. Empty / pre-versioned stored is treated as `v1.0.0` so the first bump above 1.x nudges them.
- `internal/refresh/run.go` — `OnBump(opts)` scans sidecars, formats per-tool nudge with the right `activate <tool>` command.
- `cmd/common/root.go` — when `versioncheck.RunOnce` returns Bump drift, dispatch `refresh.OnBump`. versioncheck classifies, root composes.
- `cmd/{gradle,bazel,xcode,ccache}/activate_*.go` — write/refresh the per-tool sidecar (or existing config) on successful activate.

Retired the registry + `Mark` / `lock` machinery — sidecars next to each tool's real config are the canonical source.

## Why sidecars vs. a central registry

- Sidecars survive partial wipes — deleting `~/.local/state/bitrise-build-cache` doesn't lose tracking.
- Each activate command already has a write site for its real config; piggy-backing the version field there avoids a separate file + flock.
- Reads are cheap: four small JSON files, no global lock.

## Paths package extension (bonus)

`internal/paths` now also owns the gradle + bazel sidecar paths (`BitriseCacheDir(tool)` / `BitriseCacheFile(tool, name)`), continuing the unification from #370.

## Phase 2 (deferred)

Per-tool replay handlers that programmatically re-run activate from persisted state so the user doesn't have to. Requires reconstructing activate args from existing config files + keychain + env — non-trivial and scope-bloating here. Notify text points users to the right command in the meantime.

## Test plan

- [x] `make check` passes (lint-fix + unit tests).
- [x] Scanner round-trip across the four config layouts.
- [x] `needsNudge` major-bump vs minor-bump vs empty-stored cases.
- [x] `OnBump` end-to-end.
- [ ] Manual: activate gradle locally → bump `toolconfig.GradleConfigVersion` to "2.0.0" → rerun CLI → verify the nudge appears with the gradle re-run command. Bump only minor → no nudge.


zsolt-marta-bitrise:
[//]: # ([skip ci])



## Summary
Implements per-tool config-version sidecars and major-bump refresh nudge. Detects when on-disk activate configs are from stale CLI versions and nudges users to re-run activation for major ConfigVersion bumps, while silently migrating minor/patch bumps via per-tool Migrator interface.


## Walkthrough
| File | Summary |
|------|---------|
| internal/toolconfig/base.go, internal/toolconfig/migrator.go | Added Tool enum (gradle/bazel/xcelerate/ccache), per-tool ConfigVersion constants (v1.0.0 baseline), Sample struct for scanner results, and Migrator interface for per-tool migrations |
| internal/refresh/scanner.go, internal/refresh/notify.go, internal/refresh/migrate.go, internal/refresh/run.go | Implemented refresh system: scanner walks four on-disk config locations, needsNudge() triggers on major bumps (empty treated as v1.0.0), needsMigrate() handles silent minor/patch bumps, OnBump() orchestrates scan-migrate-notify flow |
| internal/config/gradle/sidecar.go, internal/config/gradle/migrator.go | Added gradle sidecar read/write with ConfigVersion/WrittenAt and activation params, implemented SidecarMigrator for minor/patch bumps |
| internal/config/bazel/sidecar.go, internal/config/bazel/migrator.go | Added bazel sidecar read/write with ConfigVersion/WrittenAt and bazel-specific settings, implemented SidecarMigrator for minor/patch bumps |
| internal/config/xcelerate/config.go, internal/config/xcelerate/migrator.go | Replaced CLIVersion with ConfigVersion/WrittenAt fields, implemented ConfigMigrator for minor/patch bumps |
| internal/config/ccache/config.go, internal/config/ccache/migrator.go | Added ConfigVersion/WrittenAt fields to Config struct, implemented ConfigMigrator for minor/patch bumps |
| cmd/gradle/activate_gradle.go, cmd/bazel/activate_bazel.go, cmd/ccache/activate_cpp.go | Added sidecar writes after successful activation (gradle/bazel) and logger init (ccache) to persist config versions |
| cmd/common/root.go | Integrated refresh.OnBump() dispatch when versioncheck detects CLI version bump |

internal/refresh/registry.go
```diff
@@ -0,0 +1,171 @@
...
+
+	return nil
+}
+
+// Mark records (or updates) the entry for one tool. Used by the activate
+// subcommands so D1's bump detection knows what was previously configured.
+// Failures are returned but callers MUST treat them as advisory — failing
+// to write the registry should not fail an otherwise-successful activate.
+//
+// Concurrency: Mark is read-modify-write, so two parallel CLI invocations
```
zsolt-marta-bitrise: Is this a real use case for local? I'd try to avoid unnecessary complexity.
If there is a valid concurrency concern for local, we should guard all config file changes btw, not just the activation

internal/refresh/registry.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	defer unlock()
+
+	reg, err := Load(home)
+	if err != nil {
+		return err
+	}
+
+	reg.Entries[tool] = Entry{
```
zsolt-marta-bitrise: Why is it not enough to get the activation configs for each tool? Don't we already save configs (e.g. auth, cli version, options) for each tool?

internal/refresh/notify.go
```diff
@@ -0,0 +1,49 @@
...
+//
+// Returns silently when entries is empty (no tools previously configured)
+// or when logger is nil.
+func Notify(logger log.Logger, previousVersion, currentVersion string, entries []Entry) {
+	if logger == nil || len(entries) == 0 {
+		return
+	}
+
+	logger.Warnf(
+		"Bitrise Build Cache CLI bumped from %s to %s. Your previously-configured build-tool configs may be out of date. Re-run the matching command(s) below to refresh:",
```
zsolt-marta-bitrise: I'd suggest to

- use the per-tool configs as source of truth (see my other comment)
- store config versions there (semver)
- only notify when there is a config version upgrade in the given tool
- automatically migrate if supported (config version would be non-major version change). Add support for migrations in configs
- only notify if breaking change (major), in that case also block auto upgrade (or rather always block auto upgrade on major CLI version)
- This way we can release CLI minor versions with config major version changes without problems

internal/refresh/notify.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	logger.Warnf("Bitrise Build Cache config schema major bump — re-run the matching activate command(s):")
+	for _, s := range stale {
+		logger.Warnf("  • %s   # config %s → %s",
+			activateCommand(s.Tool), displayVersion(s.ConfigVersion), currents[s.Tool])
+	}
+}
+
+// needsNudge reports whether stored MAJOR < current MAJOR.
```
zsolt-marta-bitrise: [bitrise-plugin-ai-reviewer]: internal/refresh/notify.go:71:0:4cfb3359

[//]: # ([skip ci])


**📋 Guideline**

The comment states that empty stored is treated as `v0.0.0`, but the code actually treats it as `v1.0.0` (see line 81). The comment should be updated to match the actual implementation.

🔄 Suggestion:
```suggestion
// needsNudge reports whether stored MAJOR < current MAJOR.
// Empty stored is treated as `v1.0.0` so the very first major bump (1.0.0 →
// 2.0.0) leaves pre-versioned configs alone but a future 2.0.0 nudges them.
func needsNudge(stored, current string) bool {
```

zsolt-marta-bitrise (commented):

internal/config/gradle/sidecar.go
```diff
@@ -0,0 +1,111 @@
...
+	// CacheEnabled mirrors the --cache flag at activate time.
+	CacheEnabled bool `json:"cacheEnabled,omitempty"`
+	// CachePushEnabled mirrors the --cache-push flag.
+	CachePushEnabled bool `json:"cachePushEnabled,omitempty"`
+	// AnalyticsEnabled mirrors the --analytics flag.
+	AnalyticsEnabled bool `json:"analyticsEnabled,omitempty"`
+}
+
+const (
+	sidecarDirRelative = ".bitrise/cache/gradle"
```
zsolt-marta-bitrise: Please use the paths package for these too, to have a unified list of files touched by the CLI (applies to other tools too)

internal/refresh/notify.go
```diff
@@ -0,0 +1,106 @@
...
+	currents := CurrentConfigVersions()
+
+	var stale []toolconfig.Sample
+	for _, s := range samples {
+		want, ok := currents[s.Tool]
+		if !ok {
+			continue
+		}
+
+		if needsNudge(s.ConfigVersion, want) {
```
zsolt-marta-bitrise, line 55: For smaller (minor/patch) changes we'll need the auto-migrate I mentioned earlier. Let's add a common interface for the sidecars that allow the automigration

internal/refresh/notify.go
Badlazzor: Fixed in 1995fdd — the doc on `needsNudge` now matches the v1.0.0 fallback in the code.

internal/config/gradle/sidecar.go
Badlazzor: Done in f8aa16e (sidecars) and the earlier f4fdb1b — `internal/paths` now owns the gradle / bazel sidecar dirs (`BitriseCacheDir(tool)`/`BitriseCacheFile(tool, name)`), plus the xcelerate / ccache cache locations from the prior cascade.

internal/refresh/notify.go
Badlazzor: Added in 0066300 — new `toolconfig.Migrator` interface with per-tool implementations (gradle / bazel sidecar, xcelerate / ccache config). `refresh.OnBump` now runs `Migrate` for minor / patch bumps (silent re-stamp) and `Notify` for major bumps (nudge). All four tools currently sit at ConfigVersion 1.0.0 so no real migration path runs yet — the contract is in place for when a per-tool ConfigVersion gains a real minor bump.

internal/config/ccache/migrator.go
```diff
@@ -0,0 +1,44 @@
...
+	cfg, err := ReadConfig(osProxy, utils.DefaultDecoderFactory{})
+	if err != nil {
+		if errors.Is(err, fs.ErrNotExist) {
+			return nil
+		}
+
+		return fmt.Errorf("read ccache config: %w", err)
+	}
+
+	cfg.CLIVersion = common.GetCLIVersion(m.Logger)
```
zsolt-marta-bitrise: Note: this will lead to 

- either having to bump config version on CLI version (not great)
- accepting that we'll have an older CLI version in the config. The CLI will auto-update but the config migration will not happen because the config version is the same

Why do we have the CLI version in the config? Can't we use the CLI binary's version everywhere? That can be the source of truth

internal/config/xcelerate/migrator.go
```diff
@@ -0,0 +1,44 @@
...
+	cfg, err := ReadConfig(osProxy, utils.DefaultDecoderFactory{})
+	if err != nil {
+		if errors.Is(err, fs.ErrNotExist) {
+			return nil
+		}
+
+		return fmt.Errorf("read xcelerate config: %w", err)
+	}
+
+	cfg.CLIVersion = common.GetCLIVersion(m.Logger)
```
zsolt-marta-bitrise: Same as https://github.com/bitrise-io/bitrise-build-cache-cli/pull/362/changes#r3436988422

internal/config/ccache/migrator.go
zsolt-marta-bitrise: Actually, this applies to the sidecars too https://github.com/bitrise-io/bitrise-build-cache-cli/pull/362/changes#diff-971d3dfcc4fe78424c8c44bde36428c3d152af595514957a5a1346bbbf724a26R20

Probably we need the CLI version to be able to report to the analytics backends

We can just automatically bump the CLI version in all cases (even if `!needsMigration()` and note that in the migration. This aligns with the original purpose that CLI version is independent of the config version

Badlazzor: I think the version was a leftover, that was only tripping us up. Removed it from each config/sidecar, all tests (unit and CI) still green, so that matches with my expectation.

zsolt-marta-bitrise: Also checked the DB, Gradle doesn't have a CLI version field, for xcode we're still saving it 

```
[Bitrise Analytics] Sending invocation data to: https://xcode-analytics.services.bitrise.io/invocations/590d8d72-7185-4239-90c4-e9c58cef2a9a
[Bitrise Analytics] Payload: 
...
"cliVersion":"v2.0.0-20260619073935-9fa9bcf24116+dirty"
```

zsolt-marta-bitrise (approved):

internal/refresh/scanner.go
```diff
@@ -0,0 +1,91 @@
...
+	return toolconfig.Sample{
+		Tool:          toolconfig.Bazel,
+		ConfigVersion: s.ConfigVersion,
+		WrittenAt:     s.WrittenAt,
+		ConfigPath:    bazelconfig.SidecarFilePath(home),
+	}, true
+}
+
+func scanXcelerate(home string) (toolconfig.Sample, bool) {
+	osProxy := utils.DefaultOsProxy{}
```
zsolt-marta-bitrise, line 64: [bitrise-plugin-ai-reviewer]: internal/refresh/scanner.go:64:0:721744f1

[//]: # ([skip ci])


**🐛 Bug**

The `home` parameter is not used by this scanner function. Instead, it creates a new `osProxy` and relies on it to determine the home directory. This is inconsistent with `scanGradle` and `scanBazel`, which use the `home` parameter directly. If the `home` parameter differs from the system home directory (e.g., in tests), the scanner will read from the wrong location. The xcelerate `ReadConfig` function should accept a `home` parameter, or the scanner should pass a custom `osProxy` that uses the provided `home` value.

<details>

```
In internal/refresh/scanner.go at line 64, Modify scanXcelerate to use the home parameter consistently with scanGradle and scanBazel. Either update xcelerateconfig.ReadConfig to accept a home parameter, or create an osProxy that uses the provided home value.
```

</details>

internal/config/xcelerate/migrator.go
```diff
@@ -0,0 +1,42 @@
...
+)
+
+// ConfigMigrator implements toolconfig.Migrator for ~/.bitrise-xcelerate/config.json.
+type ConfigMigrator struct {
+	Logger log.Logger
+}
+
+func (ConfigMigrator) Tool() toolconfig.Tool { return toolconfig.Xcelerate }
+
+func (m ConfigMigrator) Migrate(_ string) error {
```
zsolt-marta-bitrise, line 22: [bitrise-plugin-ai-reviewer]: internal/config/xcelerate/migrator.go:22:0:00663009

[//]: # ([skip ci])


**🐛 Bug**

The `home` parameter (named `_` to indicate it's unused) is not used by this migrator. Instead, it creates a new `osProxy` and relies on it to determine the home directory. This is inconsistent with the gradle `SidecarMigrator`, which uses the `home` parameter directly. If the `home` parameter differs from the system home directory (e.g., in tests), the migrator will read from and write to the wrong location.

<details>

```
In internal/config/xcelerate/migrator.go at line 22, Modify the xcelerate ConfigMigrator to use the home parameter consistently with the gradle SidecarMigrator. Either update ReadConfig to accept a home parameter, or create an osProxy that uses the provided home value.
```

</details>

zsolt-marta-bitrise (commented):

PR #361, opened 2026-06-11
feat: ACI-5037 version-drift detector + --no-update-check flag

Badlazzor:
**Stacked on #360 (B4). GitHub will auto-retarget down the chain as parents merge.**

## Summary

- Persists the running CLI version under \`~/.local/state/bitrise-build-cache/version-state.json\` on every invocation.
- On the next run, classifies drift as \`FirstRun\` / \`NoChange\` / \`Bump\` — D3 (config auto-refresh, ACI-5039) will consume the Bump signal.
- When not on CI and \`--no-update-check\` isn't set, queries the GitHub latest-release endpoint (2s timeout, throttled to once per 4h via persisted \`LastNudgeAt\`) and writes a one-line nudge to stderr when the installed binary is behind. Stderr (not stdout) — keeps JSON-parsing stdout consumers clean.

## Architecture

- \`internal/versioncheck/state.go\` — atomic write-temp-then-rename of the JSON state file.
- \`internal/versioncheck/detector.go\` — pure \`Detect(state, currentVersion) DriftResult\`.
- \`internal/versioncheck/nudge.go\` — \`ShouldNudge\` (typed sentinel \`ErrNudgeSuppressed\`), \`FetchLatestVersion\`, \`IsBehind\`.
- \`internal/versioncheck/run.go\` — \`Run\` / \`RunOnce\` orchestrator; \`sync.Once\` guard so multiple PreRun re-entries don't double-fire.
- Hooked into \`cmd/common/root.go\` \`PersistentPreRun\`. Skipped for \`version\`, \`help\`, \`completion\` subcommands so their output stays deterministic.

## Best-effort guarantees

- Network failures never block the CLI.
- \`LastNudgeAt\` is NOT advanced on network error — the next run retries.
- \`ErrNudgeSuppressed\` is swallowed by \`Run\` (typed sentinel, \`//nolint:nilerr\` annotated with reason).

## CI

No new e2e — the network nudge would either need to hit real api.github.com (flaky / rate-limited) or run a fake server with plumbing through Bitrise VMs. Unit tests cover the behaviour via \`httptest\`: cooldown, \`--no-update-check\`, \`IsCI\` suppression, network error path, v-prefix tolerance, devel-build no-nudge case.

## Follow-ups

- ACI-5039 (D3) consumes the \`Bump\` signal to re-run \`activate\` for previously-configured build tools.
- ACI-5038 (D2) ships the actual \`update\` subcommand the nudge text mentions.

## Test plan

[...]

PR #365, opened 2026-06-12
[ACI-5084] Add `login` for local browser OAuth authentication

imrekel:
## What

Adds `bitrise-build-cache auth login` — a browser OAuth sign-in for **local** build-cache use:

```sh
bitrise-build-cache auth login     # browser → pick a workspace → done
bitrise-build-cache auth logout
```

It mints a Bitrise PAT, lets you pick a workspace, stores both **in the OS keychain**, and **auto-refreshes** the PAT before cache calls — so locally you no longer have to create a PAT by hand and export `BITRISE_BUILD_CACHE_AUTH_TOKEN` + `BITRISE_BUILD_CACHE_WORKSPACE_ID`.

The OAuth machinery is ported from the bitrise-cli implementation (same WorkOS AuthKit flow), reusing its WorkOS client identity — **no new WorkOS/monolith setup needed**.

> Targets `aci-4337-cli`, not `main` — it builds on the keychain auth store (#350) and the `charmbracelet/huh` interactive UI (#351) from the CLI epic.

## CI and manual paths are unchanged

The stored login is the **lowest-priority source**. `ResolveAuthConfig` precedence (the first two unchanged):
1. `BITRISE_BUILD_CACHE_AUTH_TOKEN` (+ `BITRISE_BUILD_CACHE_WORKSPACE_ID`) — manual, wins.
2. `BITRISEIO_BITRISE_SERVICES_ACCESS_TOKEN` (the service JWT auto-set on Bitrise CI) — wins.
3. **OS keychain** — a manual `auth set` credential *or* the new OAuth login.
4. multiplatform config on disk.

The OAuth login lands in the **same keychain slot** the rest of the CLI already resolves, so `auth list` / `auth token` and every `ResolveAuthConfig` call site see it with no per-command changes. A `PersistentPreRun` hook refreshes the stored PAT **in place in the keychain** when it's stale (so the keychain tier then serves a live PAT); it's skipped when an env credential is set, so **refresh never runs on CI**. `login` also **warns** if `BITRISE_BUILD_CACHE_AUTH_TOKEN` / the service JWT would shadow the login you just saved.

## How

- **`internal/oauth/`** — PKCE + loopback authorization-code flow (RFC 8252), `code→JWT` (WorkOS) and `JWT→PAT` (monolith `/oidc/token`, RFC 8693) exchanges, and the `EnsureFresh` refresh ladder. Persists via the OS keychain: `Load`/`Save`/`Clear` map to/from `keychain.Credentials`, which gained optional OAuth fields (PAT/JWT/refresh-token + expiries) alongside `auth_token`/`workspace_id`. Identity defaults are env-overridable (`BITRISE_OAUTH_ISSUER` / `BITRISE_OIDC_TOKEN_ENDPOINT` / `BITRISE_OAUTH_CLIENT_ID`). stdlib-only.
- **`internal/bitriseapi/`** — minimal `GET /organizations` client (`Authorization: token <PAT>`) for the workspace picker.
- **`cmd/common/login.go`** — `auth login` / `auth logout` (registered under the `auth` command, alongside set/list/clear/token); `auth logout` only clears an OAuth-managed credential (a manual `auth set` is left for `auth clear`). The workspace picker uses **`charmbracelet/huh`** (`huh.NewSelect`, the same library as `activate --interactive`) on a TTY, with a **numbered-prompt fallback** for non-TTY / CI / tests. `--workspace <slug>` skips the picker.
- **`status`** appends a read-only `Auth:` line and includes an `auth` object in `--json` (source / workspace / token expiry; never the token). Source is reported off `ResolveAuthConfig`: env PAT, CI JWT, OS keychain, OAuth login (with expiry), or multiplatform config.

## Testing

`-tags unit` tests cover: PKCE, the loopback callback, each token call, the full `EnsureFresh` refresh ladder (incl. refresh-token rotation), the keychain-backed credential store round-trip (via the go-keyring in-memory mock), the workspace API client, shadow-token detection, and the read-only auth-status (env / JWT / keychain / OAuth / multiplatform). `go build` + `make lint` (0 issues) + `make test-unit` green. (Note: a pre-existing `cmd/xcode` xcelerate-proxy test hangs locally on macOS — confirmed identical on a clean `aci-4337-cli` checkout, unrelated to this PR.)

## Open items to confirm before relying on it
- **Workspace ID format:** the picker stores the org **slug**; confirm that's what the cache's `x-org-id` / `BITRISE_BUILD_CACHE_WORKSPACE_ID` expects (vs a UUID). `--workspace` overrides regardless.
- **Short-lived PAT for config-baking commands:** the OAuth PAT is ~1h. For direct-gRPC commands (save/restore) the per-invocation refresh handles it. For `activate gradle`/`bazel`/`enable-for`, which **bake the token into generated config** (gradle.properties/.bazelrc), a 1h PAT expires until `activate` is re-run — worth a longer CLI PAT lifetime or a credential-helper approach as a follow-up.


zsolt-marta-bitrise:
[//]: # ([skip ci])



## Summary
Adds `bitrise-build-cache auth login` — browser-based OAuth sign-in for local build-cache use. Mints a Bitrise PAT, lets users pick a workspace, stores both in the OS keychain, and auto-refreshes the PAT before cache calls. Renames `auth list` to `auth status` and adds auth credential source/expiry reporting.


## Walkthrough
| File | Summary |
|------|---------|
| internal/oauth/ | New OAuth package: PKCE + loopback authorization-code flow, token exchanges (code→JWT, JWT→PAT), EnsureFresh refresh ladder, and keychain persistence via Load/Save/Clear. |
| internal/bitriseapi/organizations.go | New minimal Bitrise REST API client for workspace listing during OAuth login. |
| internal/config/common/authsource.go | New auth source description utilities for consistent credential source/expiry reporting across status, auth status, and doctor commands. |
| internal/auth/keychain/keychain.go | Extended Credentials struct with OAuth fields (PAT/JWT/refresh-token + expiries) and IsOAuthManaged() method. |
| cmd/common/login.go, select.go | New auth login/logout commands with huh-based workspace picker and numbered-prompt fallback; hydrateStoredAuth refreshes stored OAuth credentials. |
| cmd/common/root.go | Added PersistentPreRun hook to refresh OAuth logins before command execution (skipped for login, logout, status, completion, help). |
| cmd/common/status.go | Added auth status reporting with credential source and expiry info in text and JSON output. |
| cmd/auth/auth.go | Renamed `auth list` to `auth status`; registered LoginCmd/LogoutCmd; enhanced keychain display with OAuth expiry. |
| internal/doctor/check_auth.go | Refactored to use shared DescribeResolved functions for consistent auth status formatting. |
| README.md | Documented new `auth login` feature, auto-refresh capability, and auth source precedence rules. |

internal/oauth/store.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+// IsOAuthManaged reports whether a stored credential came from the OAuth flow
+// (and can be refreshed). The refresh token is the marker.
+func (c Credentials) IsOAuthManaged() bool {
+	return c.RefreshToken != ""
+}
+
+// CredentialsPath returns the absolute path to the OAuth credential file.
+func CredentialsPath() (string, error) {
```
zsolt-marta-bitrise: Please use https://github.com/bitrise-io/bitrise-build-cache-cli/pull/361/changes#diff-21138bbccef251d04d43f4112fdf366081a50977475675cade0e68a9a2c9e86a 
once that is merged

internal/oauth/flow.go
```diff
@@ -0,0 +1,171 @@
...
+// refreshing without a browser when needed:
+//
+//	PAT valid             → return it
+//	PAT expired           → exchange JWT → new PAT
+//	+ JWT expired      → refresh-token grant → new JWT → new PAT
+//	refresh token rejected → ErrLoginRequired
+//
+// Returns ErrNotLoggedIn when no OAuth credential is stored. Persists any new
+// tokens back to disk.
+func (c Config) EnsureFresh(ctx context.Context) (Credentials, error) {
```
zsolt-marta-bitrise, line 114: Let's add some debug logging in this method to be able to see what's happening

zsolt-marta-bitrise: In some cases, like the token exchange should be actually an info level so that users can see that they are still logged in and it's working

cmd/common/login.go
```diff
@@ -0,0 +1,168 @@
...
+// any error — including "not logged in" — is ignored, leaving the command to
+// surface its own "no auth configured" error. Never overrides env/CI creds, and
+// only the local-login fallback ever does a network refresh.
+func hydrateStoredAuth(ctx context.Context) {
+	if os.Getenv("BITRISE_BUILD_CACHE_AUTH_TOKEN") != "" ||
+		os.Getenv("BITRISEIO_BITRISE_SERVICES_ACCESS_TOKEN") != "" {
+		return
+	}
+	creds, err := oauth.NewConfigFromEnv(utils.AllEnvs()).EnsureFresh(ctx)
+	if err != nil || creds.PAT == "" || creds.WorkspaceID == "" {
```
zsolt-marta-bitrise: Missing error log?

cmd/common/login.go
```diff
@@ -0,0 +1,168 @@
...
+// hydrateStoredAuth makes a prior `login` take effect for every cache command:
+// when neither a manual BITRISE_BUILD_CACHE_AUTH_TOKEN nor the CI service JWT is
+// present, it refreshes the stored OAuth PAT and exports it (plus the chosen
+// workspace) into the environment, so the existing env-based auth resolution
+// (common.ReadAuthConfigFromEnvironments) picks it up unchanged. Best-effort:
+// any error — including "not logged in" — is ignored, leaving the command to
+// surface its own "no auth configured" error. Never overrides env/CI creds, and
+// only the local-login fallback ever does a network refresh.
+func hydrateStoredAuth(ctx context.Context) {
+	if os.Getenv("BITRISE_BUILD_CACHE_AUTH_TOKEN") != "" ||
```
zsolt-marta-bitrise: I'm tryign to avoid having these envs defined in many places, we have `auth.go` already with `ReadAuthConfigFromEnvironments`. Can't we use that?

cmd/common/login.go
```diff
@@ -0,0 +1,168 @@
...
+		os.Getenv("BITRISEIO_BITRISE_SERVICES_ACCESS_TOKEN") != "" {
+		return
+	}
+	creds, err := oauth.NewConfigFromEnv(utils.AllEnvs()).EnsureFresh(ctx)
+	if err != nil || creds.PAT == "" || creds.WorkspaceID == "" {
+		return
+	}
+	_ = os.Setenv("BITRISE_BUILD_CACHE_AUTH_TOKEN", creds.PAT)
```
zsolt-marta-bitrise: We save the token in many places (shell rc, github actions files, envman, etc). Please take a look at envexport package that is being used in other places and use that here too (through an interface).

cmd/common/select.go
```diff
@@ -0,0 +1,259 @@
...
+	// errRawUnsupported means the terminal can't enter raw mode; the caller
+	// should fall back to the plain numbered prompt.
+	errRawUnsupported = errors.New("interactive selection not supported on this terminal")
+)
+
+// selectFromList lets the user pick one of items. On a real terminal it shows
+// an interactive list (↑/↓ to move, type a number to jump, Enter to select);
+// otherwise — or if raw mode is unavailable — it falls back to a plain numbered
+// prompt. Returns the chosen 0-based index.
+func selectFromList(cmd *cobra.Command, prompt string, items []string) (int, error) {
```
zsolt-marta-bitrise: Are there no libraries for this? Do we have to manually implement and maintain this? (I'm skipping the review for it for now)

cmd/common/status.go
```diff
@@ -112,6 +121,70 @@ func runStatusFeature(out, errOut io.Writer, checker *status.Checker) error {
...
+	WorkspaceID string
+	TokenExpiry string
+	Expired     bool
+}
+
+// currentAuthStatus reports which credential build-cache commands would use, in
+// the same precedence as hydrateStoredAuth: a manual env token, then the CI
+// service token, then a stored OAuth login. It never refreshes or writes.
+func currentAuthStatus() authStatusInfo {
+	if os.Getenv("BITRISE_BUILD_CACHE_AUTH_TOKEN") != "" {
```
zsolt-marta-bitrise: Please use the config common package here too to avoid duplicating these envs. You can extend that to report the source of the token too

cmd/common/root.go
```diff
@@ -30,6 +30,18 @@ In case of Bazel it's done via creating or modifying $HOME/.bazelrc.`,
 		}
 
 		configcommon.LogCLIVersion(log.NewLogger(log.WithDebugLog(IsDebugLogMode)))
+
+		// For cache commands, make a prior `bitrise-build-cache login` take
+		// effect: refresh the stored OAuth PAT and export it as the auth env
+		// vars when none are already set. No-op for login/logout themselves,
+		// and never overrides manual/CI credentials.
+		switch cmd.Name() {
+		case "login", "logout", "completion", "help", "status":
```
zsolt-marta-bitrise, line 44: Also version

cmd/common/root.go
```diff
@@ -30,6 +30,18 @@ In case of Bazel it's done via creating or modifying $HOME/.bazelrc.`,
 		}
 
 		configcommon.LogCLIVersion(log.NewLogger(log.WithDebugLog(IsDebugLogMode)))
+
+		// For cache commands, make a prior `bitrise-build-cache login` take
```
zsolt-marta-bitrise: This comment is way too verbose for a trivial code. Please check other comments too (I'm not going to comment on them now) to avoid bloating the code with unnecessary textual information. Code should be done in a way that documents itself and text is only needed to add extra justification or deeper insights. Claude unfortunately started to be extremely verbose nowadays and has to be told to cut back. This is the prompt I use in my CLAUDE.md:

```
# Comments in code

Avoid commenting obvious code flows. Avoid adding long comments, be precise and include only what's not obvious from the surrounding context. Avoid bloating existing code with new comments.
Do not describe code flows that can be read from the code itself. Do not document messages (or error strings), they would be duplications. Do not include ticket numbers unless it was a non-trivial production bug. Do not include caller sources, those are discoverable. Variable comments should be max 1 short sentence. Method comments should be max 2-3 sentences. Do not comment inside if/else flows, unless they need strong justification and not obvious from the code.
```

internal/oauth/flow.go
```diff
@@ -0,0 +1,171 @@
...
+
+	cs, err := newCallbackServer(ctx, state)
+	if err != nil {
+		return Credentials{}, err
+	}
+	defer cs.close() //nolint:contextcheck // close uses a fresh short timeout for cleanup, independent of the (possibly cancelled) login context
+	cs.start()
+
+	authURL := c.authorizeURL(challenge, state, cs.redirectURI())
+	if _, err := fmt.Fprintf(stderr, "Opening your browser to sign in to Bitrise.\nIf it doesn't open automatically, visit:\n\n  %s\n\n", authURL); err != nil {
```
zsolt-marta-bitrise: Please use logger interface we use everywhere in the CLI (applies to the whole PR)

internal/oauth/flow.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	waitCtx, cancel := context.WithTimeout(ctx, loginTimeout)
+	defer cancel()
+	code, err := cs.wait(waitCtx)
+	if err != nil {
+		return Credentials{}, err
+	}
+
+	jwtResp, err := c.exchangeCodeForJWT(ctx, code, verifier, cs.redirectURI())
```
zsolt-marta-bitrise, line 65: Let's add debug logs in this method too, and an info log on success (Error on failure)

internal/oauth/browser.go
```diff
@@ -0,0 +1,30 @@
...
+// with a background context (the launched browser must outlive any request
+// context), satisfying the noctx linter without killing the browser on cancel.
+func OpenBrowser(url string) error {
+	var cmd *exec.Cmd
+	switch runtime.GOOS {
+	case "darwin":
+		cmd = exec.CommandContext(context.Background(), "open", url) //nolint:gosec // G204: fixed command, URL from our own authorize endpoint
+	case "linux":
+		cmd = exec.CommandContext(context.Background(), "xdg-open", url) //nolint:gosec // G204: fixed command, URL from our own authorize endpoint
+	case "windows":
```
zsolt-marta-bitrise: nit: we don't have windows support in other places, doesn't make sense to add it here

internal/oauth/store.go
```diff
@@ -0,0 +1,105 @@
...
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"time"
+)
+
+// Credentials is the on-disk shape of the OAuth credential file
```
zsolt-marta-bitrise: Now we have multiple places we store auth config. Can we merge them so that auth configs are handled in a singular place? This package would still be responsible for oauth flow and refresh, but storing the creds and reading from envs (see my other comments) would be handled by a central auth package.

cmd/common/status.go
```diff
@@ -69,10 +72,16 @@ func runStatus(out, errOut io.Writer, checker *status.Checker) error {
 
 	s := checker.Status()
 	if statusJSONOutput {
+		// The --json shape is a stable contract for step integrations
```
zsolt-marta-bitrise: Why not include auth status in the json too?

cmd/common/status.go
```diff
@@ -112,6 +121,70 @@ func runStatusFeature(out, errOut io.Writer, checker *status.Checker) error {
...
+
+		return info
+	}
+
+	return authStatusInfo{Source: "none"}
+}
+
+// writeAuthLine appends a one-line auth summary to the text status output.
+func writeAuthLine(out io.Writer, a authStatusInfo) error {
+	line := "Auth: "
```
zsolt-marta-bitrise, line 180: Let's log the type of login (PAT+wsID, JWT, or oauth)

cmd/common/select.go
```diff
@@ -0,0 +1,259 @@
...
+}
+
+// decodeKeys splits a burst of raw bytes (one or more keypresses) into events.
+// Arrow keys arrive as a 3-byte escape sequence (ESC [ A/B); everything else is
+// decoded byte by byte so fast/batched input isn't dropped.
+func decodeKeys(b []byte) []keyEvent {
+	var out []keyEvent
+	for i := 0; i < len(b); i++ {
+		if b[i] == 0x1b {
+			if i+2 < len(b) && b[i+1] == '[' {
```
zsolt-marta-bitrise: [bitrise-plugin-ai-reviewer]: cmd/common/select.go:90:0:4655da91

[//]: # ([skip ci])


**🐛 Bug**

When an unknown escape sequence like ESC [ C is encountered (where C is not A or B), the code falls through the switch statement and treats it as a lone Esc (keyCancel). However, it only increments i by 0, so the next iteration will process the '[' and 'C' characters separately as keyOther events. This results in three events (keyCancel, keyOther, keyOther) instead of one (keyOther for the unknown sequence). The switch statement should have a default case to handle unknown escape sequences by skipping the entire 3-byte sequence.

🔄 Suggestion:
```suggestion
			if i+2 < len(b) && b[i+1] == '[' {
				switch b[i+2] {
				case 'A':
					out = append(out, keyEvent{kind: keyUp})
					i += 2

					continue
				case 'B':
					out = append(out, keyEvent{kind: keyDown})
					i += 2

					continue
				default:
					// Unknown escape sequence, treat as keyOther
					out = append(out, keyEvent{kind: keyOther})
					i += 2

					continue
				}
			}
```

zsolt-marta-bitrise (commented):

internal/oauth/flow.go
imrekel: Added debug logging through the `EnsureFresh` ladder, plus an info-level log when the PAT is refreshed/exchanged so users can see the login is still working. Logging goes through the go-utils `log.Logger` interface via `Config.Logger`.

cmd/common/login.go
imrekel: Added — `hydrateStoredAuth` now logs the skip reason at debug level (`logger.Debugf("OAuth login not applied: %s", err)`).

cmd/common/root.go
imrekel: `version` is handled earlier in `PersistentPreRun` — it returns before the hydrate step runs, so it never reaches this skip set. Listing it here would also trip goconst (the literal would appear 3×). Left as-is.

cmd/common/root.go
imrekel: Trimmed the verbose comments here and across the PR (oauth.go, callback.go, flow.go) down to the non-obvious bits.

internal/oauth/flow.go
imrekel: Done across the PR — the OAuth flow now logs through the go-utils `log.Logger` interface, injected via `Config.Logger` (nil-safe `debugf`/`infof`/`warnf` helpers). No more passed-in `io.Writer`.

internal/oauth/flow.go
imrekel: Added debug logs through `Login`, an info log on success, and a warning on the recoverable browser-open failure.

internal/oauth/browser.go
imrekel: Removed the windows branch — darwin/linux only now, matching the rest of the CLI.

cmd/common/status.go
imrekel: Done — `--json` now emits an `auth` object (configured / source / workspace_id / token_expiry / expired) alongside the feature flags.

cmd/common/status.go
imrekel: Done — the auth source now reports the credential type: `PAT + workspace ID (env)`, `JWT (CI service token)`, or `OAuth login`, in both the text line and the JSON `source` field.

cmd/common/select.go
imrekel: Fixed — added a `default` case in the escape-sequence decoder that emits `keyOther` for unrecognized sequences instead of falling through to `keyCancel`.

internal/oauth/store.go
imrekel: #361 is closed - what about the change?

cmd/common/login.go
imrekel: Agreed the env names shouldn't be redefined. Folding these reads into config-common `auth.go` (`ReadAuthConfigFromEnvironments`) is part of a central-auth refactor I'd like to do as a follow-up — it also touches store.go and status.go — rather than expand this PR. Tracking it with the store-merge comment.

cmd/common/login.go
imrekel: Good pointer — persisting via the `envexport` package (shell rc / github actions / envman) is a larger change; tracking it in the same central-auth follow-up. For this PR the stored OAuth credential is a local-only fallback written to the JSON store, so it doesn't need the multi-target export yet.

cmd/common/status.go
imrekel: Source reporting is now in (reports PAT / JWT / OAuth type). The env-constant reuse to avoid duplicating env names is part of the central-auth follow-up alongside the login.go comment — leaving that part for that change.

internal/oauth/store.go
imrekel: Agreed — merging the auth stores into one place is the central-auth refactor I'm tracking as a follow-up (with the login.go env-reuse and the store.go #361 comments). Keeping this PR's store self-contained for the local-only OAuth fallback for now.

cmd/common/select.go
imrekel: It's a deliberately small hand-rolled raw-mode reader (arrow/number selection) to avoid pulling in a TUI dependency for a single picker. Happy to swap in a library if you'd prefer one — open to a suggestion.

internal/oauth/store.go
zsolt-marta-bitrise: We have the paths package collecting all the files/sockets that the CLI generates, please use that for all file paths / sockets

cmd/common/login.go
zsolt-marta-bitrise: What is the scope of that auth refactor?

cmd/common/select.go
zsolt-marta-bitrise: We already have a lib added in the interactive setup flow, let's use that

cmd/common/login.go
imrekel: Done in 75611d2. `config/common` now exports the env-var names (`EnvAuthToken` / `EnvWorkspaceID` / `EnvServiceJWT`) as the single source of truth, and `hydrateStoredAuth` / `shadowingAuthEnv` use those constants instead of re-typing the literals. `status` resolves the env credential via `ReadAuthConfigFromEnvironments`.

cmd/common/login.go
zsolt-marta-bitrise: We also have keychain now for token storage, so whatever we create here should be stored there right? Without that support the new features will not work together well

cmd/common/status.go
imrekel: Done in 75611d2. `currentAuthStatus` now reads the env credential through `config/common.ReadAuthConfigFromEnvironments` (no more re-typed env literals) and reports the source off the resolved config: PAT / JWT / OAuth login.

internal/oauth/store.go
imrekel: Done in 75611d2 now that the `paths` package from #361 is on main. `CredentialsPath` derives from `paths.Default().BuildCacheAuthFile()` (added `BuildCacheDir` / `BuildCacheAuthFile` → `~/.bitrise/build-cache/auth.json`) instead of hand-joining the path.

internal/oauth/store.go
imrekel: Partially addressed in 75611d2: the env-var names are now centralized in `config/common` and the credential file location goes through the `paths` package, so those are no longer scattered. Fully folding the on-disk OAuth credential store (PAT/JWT/refresh + expiries) into `config/common` is a larger change — that store has a different shape than `CacheAuthConfig` (the resolved cache token) — so I'd like to keep it as a follow-up rather than expand this PR. Leaving this thread open to track that.

internal/oauth/flow.go
```diff
@@ -0,0 +1,170 @@
...
+	if creds.RefreshToken == "" {
+		return Credentials{}, ErrLoginRequired
+	}
+	c.debugf("Refreshing the OAuth session")
+	refreshed, err := c.refreshJWT(ctx, creds.RefreshToken)
+	if err != nil {
+		return Credentials{}, fmt.Errorf("%w (refresh failed: %w)", ErrLoginRequired, err)
+	}
+	creds.JWT = refreshed.AccessToken
+	creds.JWTExpiry = jwtExpiry(refreshed, now)
```
zsolt-marta-bitrise, line 154: [bitrise-plugin-ai-reviewer]: internal/oauth/flow.go:154:0:4655da91

[//]: # ([skip ci])


**🐛 Bug**

The `now` variable was captured at line 124, before the refresh token exchange at line 149. By the time this line executes, significant time may have elapsed (network latency, etc.), causing the JWT expiry to be calculated from a stale timestamp. This results in an expiry time that's further in the future than it should be, potentially delaying the next refresh.

🔄 Suggestion:
```suggestion
	creds.JWT = refreshed.AccessToken
	creds.JWTExpiry = jwtExpiry(refreshed, time.Now())
	if refreshed.RefreshToken != "" { // WorkOS may rotate the refresh token
		creds.RefreshToken = refreshed.RefreshToken
	}
```

zsolt-marta-bitrise (commented):

cmd/common/login.go
imrekel: Scope + done (a254cba): I retargeted this PR onto `aci-4337-cli` and the OAuth login now stores its credential in the OS keychain (the canonical store) instead of a separate file. `internal/oauth` Load/Save/Clear delegate to `internal/auth/keychain`; `keychain.Credentials` gained optional OAuth fields (refresh token + PAT/JWT expiries) so the refresh ladder works. Because it lives in the keychain, `ResolveAuthConfig`'s keychain tier serves it to every command and `auth list` / `auth token` see it. No `envexport` is needed here: the credential isn't baked into shell rc / GitHub Actions / envman — it's resolved at runtime from the keychain, and `hydrateStoredAuth` just refreshes it in place when the PAT is stale (skipped when an env credential is set, so CI never refreshes).

cmd/common/select.go
imrekel: Done in a254cba — the workspace picker now uses `charmbracelet/huh` (`huh.NewSelect`), the same library as `activate --interactive`. Deleted the hand-rolled raw-mode key reader and its tests; a numbered prompt remains as the non-TTY/CI fallback.

internal/oauth/store.go
imrekel: Now fully addressed in a254cba (PR retargeted onto `aci-4337-cli`): the OAuth login is stored in the OS keychain via `internal/auth/keychain`, so there's a single auth store shared with `auth set`. The bespoke `~/.bitrise/build-cache/auth.json` file is gone, and `internal/oauth` is just a thin mapping over `keychain.Credentials`.

internal/oauth/store.go
imrekel: Follow-up to my earlier reply: the credential file is now removed entirely — the OAuth login is stored in the OS keychain (a254cba), so there's no longer a `~/.bitrise/build-cache` path at all.

internal/oauth/flow.go
imrekel: Fixed in 6c43aea — `jwtExpiry` is now called with a `time.Now()` captured at the refresh exchange rather than the `now` from EnsureFresh entry, so `expires_in` is added to the right base. (In practice it was conservative either way given the 60s refresh skew, but it is correct now.)

internal/oauth/flow.go
```diff
@@ -0,0 +1,170 @@
...
+
+	now := time.Now()
+
+	return Credentials{
+		PAT:          pat,
+		PATExpiry:    patExpiry,
+		JWT:          jwtResp.AccessToken,
+		JWTExpiry:    jwtExpiry(jwtResp, now),
+		RefreshToken: jwtResp.RefreshToken,
+	}, nil
```
zsolt-marta-bitrise, line 82: [bitrise-plugin-ai-reviewer]: internal/oauth/flow.go:63:83:3bfcf111

[//]: # ([skip ci])


**🐛 Bug**

The `now` timestamp is captured after the token exchanges (lines 64 and 69), but it's used to calculate JWT expiry relative to that later time. This causes the JWT to appear to expire sooner than it actually does, since the expiry duration is added to a time that's after the actual token exchange. The `now` should be captured before the exchanges, similar to how it's done in EnsureFresh at line 124.

🔄 Suggestion:
```suggestion
	c.debugf("Exchanging authorization code for a token")
	now := time.Now()
	jwtResp, err := c.exchangeCodeForJWT(ctx, code, verifier, cs.redirectURI())
	if err != nil {
		return Credentials{}, fmt.Errorf("exchange authorization code: %w", err)
	}
	c.debugf("Exchanging token for a Bitrise access token")
	pat, patExpiry, err := c.exchangeJWTForPAT(ctx, jwtResp.AccessToken)
	if err != nil {
		return Credentials{}, fmt.Errorf("exchange token for a Bitrise PAT: %w", err)
	}
	c.infof("Signed in to Bitrise.")

	return Credentials{
		PAT:          pat,
		PATExpiry:    patExpiry,
		JWT:          jwtResp.AccessToken,
		JWTExpiry:    jwtExpiry(jwtResp, now),
		RefreshToken: jwtResp.RefreshToken,
	}, nil
```

zsolt-marta-bitrise (commented):

internal/oauth/flow.go
imrekel: Fixed in 7eebc19 — `now` is now captured before `exchangeCodeForJWT`, so the JWT expiry is anchored to (just before) issuance rather than after both exchanges. Same fix as the EnsureFresh path.

cmd/common/select.go
```diff
@@ -0,0 +1,67 @@
...
+		Run(); err != nil {
+		return 0, fmt.Errorf("interactive selection: %w", err)
+	}
+
+	return choice, nil
+}
+
+// numberedPrompt prints a numbered list and reads a 1-based choice from stdin.
+// Used non-interactively (pipes) or when there's no terminal.
+func numberedPrompt(cmd *cobra.Command, prompt string, items []string) (int, error) {
```
zsolt-marta-bitrise: In the interactive flow we only support terminals that support huh, I don't see why we need to maintain two versions just here. That said, it's not that large so it can stay

cmd/common/status.go
```diff
@@ -112,6 +127,85 @@ func runStatusFeature(out, errOut io.Writer, checker *status.Checker) error {
...
+	switch {
+	case !a.Configured:
+		line += "not configured (run 'bitrise-build-cache login', or set BITRISE_BUILD_CACHE_AUTH_TOKEN + BITRISE_BUILD_CACHE_WORKSPACE_ID)"
```
zsolt-marta-bitrise: Or run the interactive setup

cmd/common/login.go
zsolt-marta-bitrise: Now that the keychain code is in place, do we still need to handle env vars as a special case here? Why not rely on the auth source directly from config common?

zsolt-marta-bitrise (commented):
**Consolidate auth-source detection — this is now the 3rd implementation**

`currentAuthStatus()` here is solid (it goes through `ResolveAuthConfig`, the same path the cache actually uses, and it's the only one that knows about the OAuth-login distinction + PAT expiry). But it's now the *third* place that detects and describes where the credential comes from:

1. `internal/doctor/check_auth.go` `authCheck()` — keychain / env / JWT / none
2. `cmd/auth/auth.go` `auth list` — its own `credSource` / `probe*` audit of all sources
3. this `currentAuthStatus()` — `ResolveAuthConfig` + the OAuth/expiry extras

They already disagree, and the disagreement is user-visible: for an OAuth keychain login, `status` will now say `OAuth login (keychain), token valid until …` while `doctor` still prints a flat `OS keychain`. The source taxonomy also lives in three hand-maintained spellings that will drift.

Could we pull the **detection + description** into one place (extend `internal/config/common`, or a small `internal/authsource`) so all three consume it:

- per-source probes (keychain / env-PAT / JWT / multiplatform) — `auth list` already has these; lift them up
- the OAuth-login-vs-manual + expiry knowledge that currently only lives here, next to the keychain probe, so `doctor` picks it up for free
- a single source → human-label / JSON-shape mapping

To be clear — I'm **not** asking to unify *precedence*. `doctor` deliberately checks keychain first (to nudge users toward migrating to the keychain), and `auth list` deliberately reports every source rather than a single winner. Those selection policies should stay per-caller; it's only the "what's in source X / how do I describe it" layer that's duplicated and worth sharing. Keeps the OAuth + expiry reporting single-sourced instead of something the other two have to re-learn later.

cmd/common/login.go
zsolt-marta-bitrise: Also see https://github.com/bitrise-io/bitrise-build-cache-cli/pull/365#pullrequestreview-4570793070

zsolt-marta-bitrise (commented):
_(duplicate — see #pullrequestreview-4570793070)_

cmd/common/select.go
zsolt-marta-bitrise: Unresolved, see https://github.com/bitrise-io/bitrise-build-cache-cli/pull/365#discussion_r3474195098

cmd/common/status.go
zsolt-marta-bitrise: See https://github.com/bitrise-io/bitrise-build-cache-cli/pull/365#pullrequestreview-4570793070

imrekel: Done as part of the auth-source consolidation in 8b42eb0 (full reply on the review thread). `currentAuthStatus` now builds from the shared `config/common` describe layer (`DescribeResolved` + `SourceLabel`), the same one `doctor` and `auth list` use.

imrekel:
Re [the auth-source consolidation review](https://github.com/bitrise-io/bitrise-build-cache-cli/pull/365#pullrequestreview-4570793070): done in 8b42eb0.

Pulled the **detection + description** layer into `internal/config/common` (`authsource.go`): `AuthDescription` + `SourceLabel` (one canonical spelling per source), `Detail()`/`Expired()`, and `DescribeResolved`/`DescribeResolvedWith`/`DescribeKeychainCredentials` — which fold in the keychain OAuth-login + PAT-expiry knowledge that previously only `status` had.

All three consumers now build from it:
- `status.currentAuthStatus`
- `doctor.authCheck` — now reports an OAuth login + expiry instead of a flat `OS keychain`, fixing the user-visible drift you flagged
- `cmd/auth list` keychain row — same

As you asked, **precedence/selection stays per-caller** (doctor keychain-first, `auth list` shows-all, `status` the resolved winner); only the "what's in source X / how to describe it" layer is shared now.

cmd/common/login.go
imrekel: Good call — done in 19515f5. `login.go` no longer hand-checks the env vars: both `hydrateStoredAuth` (skip refresh when env wins) and `shadowingAuthEnv` (the "this env var shadows your login" warning) now derive that from `ResolveAuthConfig`'s `AuthSource`. Since the OAuth login is in the keychain (below env in precedence), the resolved winner *is* the shadow. Net effect: the warning now reflects what actually wins resolution rather than mere env-var presence.

cmd/common/select.go
imrekel: Dropped in 4138bfa — the workspace picker is now huh-only (no numbered fallback), matching `activate --interactive`. `runLogin` already requires an interactive TTY before the picker runs (`--workspace` is the non-interactive path), so the fallback never actually fired.

cmd/common/select.go
imrekel: Dropped in 4138bfa — the workspace picker is now huh-only (no numbered fallback), matching `activate --interactive`. `runLogin` already requires an interactive TTY before the picker runs (`--workspace` is the non-interactive path), so the fallback never actually fired.

cmd/common/status.go
imrekel: Done in the latest push — the not-configured hint now also suggests `bitrise-build-cache activate --interactive`, matching the doctor "no credentials" hint.

internal/oauth/flow.go
```diff
@@ -0,0 +1,169 @@
...
+	c.debugf("Waiting for the browser sign-in to complete")
+	waitCtx, cancel := context.WithTimeout(ctx, loginTimeout)
+	defer cancel()
+	code, err := cs.wait(waitCtx)
+	if err != nil {
+		return Credentials{}, err
+	}
+
+	c.debugf("Exchanging authorization code for a token")
+	now := time.Now() // before the exchange, so the JWT expiry isn't pushed out by the round-trip
```
zsolt-marta-bitrise, line 64: [bitrise-plugin-ai-reviewer]: internal/oauth/flow.go:64:0:7eebc193

[//]: # ([skip ci])


**🐛 Bug**

The `now` timestamp is captured before the code→JWT exchange, but it's used to calculate JWT expiry after both the code→JWT and JWT→PAT exchanges complete. This causes the JWT expiry to be calculated relative to a time captured before the actual token exchange, making the JWT appear to expire sooner than it actually does. The comment says "before the exchange" but the intent should be to capture time relative to when the JWT is actually received.

🔄 Suggestion:
```suggestion
	jwtResp, err := c.exchangeCodeForJWT(ctx, code, verifier, cs.redirectURI())
	if err != nil {
		return Credentials{}, fmt.Errorf("exchange authorization code: %w", err)
	}
	now := time.Now() // after the exchange, so JWT expiry is relative to when we got the token
```

zsolt-marta-bitrise (commented):

zsolt-marta-bitrise (approved):
LGTM, thanks for the fixes!

Badlazzor (approved):

cmd/common/login.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	return nil
+}
+
+// shadowingAuthEnv returns the env credential that takes precedence over the
+// stored OAuth login, or "". The OAuth login lives in the keychain (below env
+// in ResolveAuthConfig's precedence), so whichever env source wins the
+// resolution is exactly what shadows the just-saved login.
+func shadowingAuthEnv() string {
```
zsolt-marta-bitrise, line 110: Still obvious + unnecessary + long  comment. Again, please remove these

zsolt-marta-bitrise: And courtesy of @Badlazzor for catching it, I didn't re-review this after I already highlighted the unecessary comments.

Can be done as a followup

imrekel: Done now in 5a63b8c (no need for a follow-up). Dropped the `shadowingAuthEnv` comment to one line and trimmed the other over-long ones added by this PR — `hydrateStoredAuth`, `currentAuthStatus`, `AuthDescription`, `oauth.Credentials`, `Detail` — to the non-obvious essentials. Thanks @Badlazzor.

internal/oauth/flow.go
imrekel: This is a false positive (and contradicts the bot's earlier comment on the same lines, which asked to capture `now` *before* the exchange — which is what the code now does). Capturing `now` just before `exchangeCodeForJWT` anchors the expiry to ~issuance time; if anything it is computed marginally *earlier* than the true expiry, which is the safe/conservative direction (we refresh slightly early, never late). The skew here is sub-second versus the 60s refreshSkew, so it has no practical effect. Leaving as-is.

cmd/common/status.go
```diff
@@ -112,6 +126,73 @@ func runStatusFeature(out, errOut io.Writer, checker *status.Checker) error {
...
+	Xcode       bool           `json:"xcode"`
+	Cpp         bool           `json:"cpp"`
+	ReactNative bool           `json:"reactNative"`
+	Auth        authStatusInfo `json:"auth"`
+}
+
+// currentAuthStatus reports the credential commands would use, via config
+// common's resolution + shared AuthDescription. Never refreshes or writes.
+func currentAuthStatus() authStatusInfo {
+	cfg, source, err := configcommon.ResolveAuthConfig(utils.AllEnvs())
```
zsolt-marta-bitrise, line 152: [bitrise-plugin-ai-reviewer]: cmd/common/status.go:151:0:a254cba9

[//]: # ([skip ci])


**🐛 Bug**

The function returns `authStatusInfo{Source: "none"}` when there's an error from ResolveAuthConfig, but this masks the actual error. If ResolveAuthConfig returns an error (e.g., invalid JWT format), the user won't know why auth is unconfigured. The error should be logged or the function should distinguish between "not configured" and "configuration error".

<details>

```
In cmd/common/status.go at line 151, Log or handle ResolveAuthConfig errors explicitly instead of silently treating them as "not configured".
```

</details>

cmd/common/login.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	return workspaces[idx].Slug, nil
+}
+
+// hydrateStoredAuth refreshes a stale keychain OAuth login so ResolveAuthConfig
+// serves a live PAT. Skipped when an env credential wins (env precedence, and CI
+// must never refresh); this is the only path that does a network refresh.
+func hydrateStoredAuth(ctx context.Context) {
+	switch _, source, _ := configcommon.ResolveAuthConfig(utils.AllEnvs()); source {
```
zsolt-marta-bitrise: [bitrise-plugin-ai-reviewer]: cmd/common/login.go:152:0:19515f55

[//]: # ([skip ci])


**🐛 Bug**

The switch statement has a logic issue: when the source is AuthSourceNone, AuthSourceKeychain, or AuthSourceMultiplatform, the code falls through and attempts to refresh. However, for AuthSourceNone (no credentials configured), calling EnsureFresh will fail with ErrNotLoggedIn. The function should only attempt refresh when source is AuthSourceKeychain (an OAuth-managed credential). For AuthSourceNone and AuthSourceMultiplatform, refresh is not applicable.

<details>

```
In cmd/common/login.go at line 152, Only attempt OAuth refresh when source is AuthSourceKeychain; skip for AuthSourceNone and AuthSourceMultiplatform.
```

</details>

zsolt-marta-bitrise (commented):

cmd/common/login.go
imrekel: Good simplification — done in e1575d8. `hydrateStoredAuth` now refreshes only when the resolved source is `AuthSourceKeychain`. A keychain OAuth login always wins over multiplatform in precedence, so `None`/`Multiplatform` can never have one to refresh; gating on keychain drops the needless keychain read + debug line on those paths. (`EnsureFresh` still self-guards via `IsOAuthManaged`, so a manual `auth set` keychain entry is a no-op.)

cmd/common/status.go
imrekel: Addressed in e1575d8. The common case is fine — `ResolveAuthConfig` returns `ErrAuthTokenNotProvided` when nothing is set, which legitimately *is* "none". But a genuine failure (e.g. a malformed `BITRISEIO_BITRISE_SERVICES_ACCESS_TOKEN`) is a different error, and that is now surfaced as `Source: "error"` + the message (in the text line and `--json`) rather than mislabeled "not configured".

internal/oauth/flow.go
```diff
@@ -0,0 +1,169 @@
...
+	if creds.RefreshToken == "" {
+		return Credentials{}, ErrLoginRequired
+	}
+	c.debugf("Refreshing the OAuth session")
+	refreshed, err := c.refreshJWT(ctx, creds.RefreshToken)
+	if err != nil {
+		return Credentials{}, fmt.Errorf("%w (refresh failed: %w)", ErrLoginRequired, err)
+	}
+	creds.JWT = refreshed.AccessToken
+	creds.JWTExpiry = jwtExpiry(refreshed, time.Now()) // relative to the exchange, not EnsureFresh entry
```
zsolt-marta-bitrise: [bitrise-plugin-ai-reviewer]: internal/oauth/flow.go:147:153:3bfcf111

[//]: # ([skip ci])


**🐛 Bug**

The JWT expiry is calculated using time.Now() called AFTER the refreshJWT network call completes. This is inconsistent with the Login function (line 64) which captures time.Now() BEFORE the exchange. The expiry should be relative to when the request was made, not when the response was received, to avoid the round-trip time pushing out the expiry.

🔄 Suggestion:
```suggestion
	c.debugf("Refreshing the OAuth session")
	now := time.Now() // before the exchange, so the JWT expiry isn't pushed out by the round-trip
	refreshed, err := c.refreshJWT(ctx, creds.RefreshToken)
	if err != nil {
		return Credentials{}, fmt.Errorf("%w (refresh failed: %w)", ErrLoginRequired, err)
	}
	creds.JWT = refreshed.AccessToken
	creds.JWTExpiry = jwtExpiry(refreshed, now)
```

zsolt-marta-bitrise (commented):

PR #367, opened 2026-06-15
docs: ACI-5043 Xcode.app integration one-pager

Badlazzor:
## Summary

- New `docs/xcode-app.md` — one-page guide for enabling the compile cache when developers press ▶ in Xcode.app. Stacks on the `xcode-app enable / disable` PR (#366).
- Cross-linked from the README's Install section so it surfaces next to the existing local-dev install pointer.

## Contents

- TL;DR (4 commands).
- How it works (override slot + LaunchAgent + proxy daemon, one paragraph each).
- Prerequisites table — what each `bitrise-build-cache activate xcode` / `daemon install + up` step is responsible for.
- Enable + relaunch step (with the GUI-session caveat that `launchctl setenv` only reaches processes spawned after it ran).
- Verification recipe — `launchctl getenv` + `doctor` + a build.
- `#include` chaining of a pre-existing `XCODE_XCCONFIG_FILE`.
- Disable behaviour.
- Repo-controlled helper script pattern (the Review-table P4 ask) — sample `scripts/enable-bitrise-build-cache.sh` + the explicit note that auth credentials stay personal and out of the repo.
- Troubleshooting table.

## Stacked PR

Based on `aci-5041-xcode-app-enable-disable`. GitHub auto-retargets to the next ancestor when #366 lands. Diff stays clean against the parent.

## Test plan

- [x] Renders correctly in GitHub's markdown preview.
- [ ] Smoke-test the TL;DR + helper-script flow on a fresh macOS box.
- [ ] Verify the troubleshooting table's suggestions actually surface the listed symptoms.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,10 @@ Authentication is via two env vars (PAT + workspace ID) — see the [post-instal
 
 > The CLI configures the environment it's running in. If you're running commands in Docker containers, run the CLI inside the same container as Gradle/Bazel/Xcode/ccache.
 
+### Xcode.app (GUI builds)
+
+Pressing ▶ in Xcode.app bypasses the `xcodebuild` wrapper. Use `bitrise-build-cache xcode-app enable` to install an `XCODE_XCCONFIG_FILE` override so the GUI build pipeline also picks up the cache. See [`docs/xcode-app.md`](docs/xcode-app.md) for the full flow + a repo-controlled helper script pattern for team-wide rollout. macOS only.
+
 
 ## What does the CLI do on a high level?
 
--- a/docs/xcode-app.md
+++ b/docs/xcode-app.md
@@ -1,0 +1,211 @@
+# Use the Bitrise Build Cache from Xcode.app (GUI builds)
+
+A one-page guide for enabling the compile cache when you press ▶ in
+Xcode.app. The default `xcodebuild` wrapper only kicks in for command-line
+builds — this page covers the GUI case.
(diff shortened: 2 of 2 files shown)


Badlazzor:
## Merge-time TODO — swap concrete verification commands back to \`doctor\`

The B1 fix replaced \`bitrise-build-cache doctor\` references with concrete commands (\`launchctl list | grep\`, \`test -S\` on the socket path, \`ls\` the proxy log dir) because \`doctor\` doesn't exist on this PR's branch ancestry. It does ship on the parallel PR #355 (\`aci-5035-doctor-subcmd\`).

When PR #355 and this PR both land on main, the docs should swap the concrete commands back to:

- \`launchctl list | grep io.bitrise.build-cache\` + \`test -S "$(jq -r .proxySocketPath ~/.bitrise-xcelerate/config.json)"\` → \`bitrise-build-cache doctor\` (single command, friendlier output)
- \`ls -lt ~/.local/state/xcelerate/logs/\` → keep as a complementary check, but the primary verification command should be \`doctor\`

Affected sections in \`docs/xcode-app.md\`:
- \`## Verify the override is in effect\` (the three-step recipe)
- \`## Troubleshooting\` (proxy-socket-missing row)

Track this here so whoever lands the second of the two PRs (or the first PR after the other merges) doesn't forget. The check is small — fine to do in a follow-up commit after both PRs are merged rather than rebasing.

PR #368, opened 2026-06-15
feat: ACI-5049 browse subcommand

Badlazzor:
## Summary

- Adds `bitrise-build-cache browse [invocation-id]` so users can jump from the terminal to the Bitrise Build Cache dashboard in one command.
- Standalone PR — no daemon / xcelerate / xcode-app stack dependency. Bases off `main`.

## Behaviour

| Invocation                                            | URL emitted                                                                              |
| ----------------------------------------------------- | ---------------------------------------------------------------------------------------- |
| `bitrise-build-cache browse`                          | `https://app.bitrise.io/build-cache/<workspace>/invocations?source=local`                |
| `bitrise-build-cache browse inv_xyz`                  | `https://app.bitrise.io/build-cache/<workspace>/invocations/inv_xyz`                     |
| `bitrise-build-cache browse --print`                  | (URL printed only; no launcher)                                                          |

- Workspace ID resolved from `BITRISE_BUILD_CACHE_WORKSPACE_ID` env var. Browse doesn't need an auth token — the URL opens in the user's browser, which handles its own session. Missing-workspace returns a sentinel `pkg/browse.ErrWorkspaceNotConfigured` (distinct from `common.ErrWorkspaceIDNotProvided`, which couples workspace + token).
- Launcher: `/usr/bin/open` on darwin, `xdg-open` on linux/BSD. Unknown platforms or a launcher-side error degrade to "URL was logged — copy it" rather than hard-failing the command.

## `source=local` note

The dashboard ignores the `source` query param today — F4 (ACI-5047) is the BE-side ticket that wires it up. Sending it now is forward-compatible and harmless on the current dashboard.

## Layout

- `cmd/browse/browse.go` — thin cobra wrapper, registered via blank import in `main.go`
- `pkg/browse/browser.go` — public `Browser` struct + `Params` + sentinel error
- `internal/browse/`:
  - `url.go` — pure URL builder, dispatched on (workspace, invocationID, sourceFilter)
  - `open.go` — opener interface + DefaultOpener (exec the platform launcher)

## Test plan

- [x] `make check` (lint + unit) passes locally.
- [x] Manual smoke:
  - `BITRISE_BUILD_CACHE_WORKSPACE_ID=ws_test go run . browse --print` → `…/build-cache/ws_test/invocations?source=local`
  - `BITRISE_BUILD_CACHE_WORKSPACE_ID=ws_test go run . browse --print inv_xyz` → `…/build-cache/ws_test/invocations/inv_xyz`
- [ ] Quick darwin smoke without `--print` — should pop the default browser.

## Out of scope

- F4 (BE accepts `source` field) — separate ticket (ACI-5047).
- Mac OS keychain workspace fallback — separate ticket (ACI-5028, on the M1 stack); env-var is the canonical source for now.

### DIFF
--- a/cmd/browse/browse.go
+++ b/cmd/browse/browse.go
@@ -1,5 +1,3 @@
-// Package browse exposes the `bitrise-build-cache browse` cobra subcommand.
-// Thin wrapper that maps flags into pkg/browse and prints the dashboard URL.
 package browse
 
 import (
--- a/cmd/browse/browse.go
+++ b/cmd/browse/browse.go
@@ -1,0 +1,69 @@
+// Package browse exposes the `bitrise-build-cache browse` cobra subcommand.
+// Thin wrapper that maps flags into pkg/browse and prints the dashboard URL.
+package browse
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/bitrise-io/go-utils/v2/log"
+	"github.com/spf13/cobra"
+
+	"github.com/bitrise-io/bitrise-build-cache-cli/v2/cmd/common"
+	"github.com/bitrise-io/bitrise-build-cache-cli/v2/internal/utils"
+	browsepkg "github.com/bitrise-io/bitrise-build-cache-cli/v2/pkg/browse"
+)
+
+// browseParams collects flag-mapped inputs for the subcommand. Kept in a
+// struct so additional flags can land without expanding the RunE
+// signature. Exported field-by-field for inspection in tests.
+//
+//nolint:gochecknoglobals // mirrors the pattern in cmd/xcode, cmd/gradle, …
+var browseParams struct {
+	printOnly bool
+}
+
+//nolint:gochecknoglobals
+var browseCmd = &cobra.Command{
+	Use:   "browse [invocation-id]",
+	Short: "Open the Bitrise Build Cache dashboard for the configured workspace",
+	Long: `browse opens the user's default browser at the Bitrise Build Cache dashboard, ` +
+		`pre-filtered to the configured workspace (BITRISE_BUILD_CACHE_WORKSPACE_ID) and ` + "`source=local`" + ` invocations. ` +
+		`Pass an optional invocation ID positional argument to deep-link to a specific invocation page. ` +
+		`Use ` + "`--print`" + ` to skip the launcher and only emit the URL — useful in headless / CI sessions.`,
+	Args:         cobra.MaximumNArgs(1),
+	SilenceUsage: true,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		logger := log.NewLogger(log.WithDebugLog(common.IsDebugLogMode))
+
+		params := browsepkg.Params{
+			Envs:      utils.AllEnvs(),
+			PrintOnly: browseParams.printOnly,
+		}
+		if len(args) == 1 {
+			params.InvocationID = args[0]
+		}
+
+		b := &browsepkg.Browser{Logger: logger}
+		if _, err := b.Open(cmd.Context(), params); err != nil {
+			if errors.Is(err, browsepkg.ErrWorkspaceNotConfigured) {
+				return err //nolint:wrapcheck // sentinel
+			}
+
+			return fmt.Errorf("browse: %w", err)
+		}
+
+		return nil
+	},
+}
+
+func init() {
+	browseCmd.Flags().BoolVar(
+		&browseParams.printOnly,
+		"print",
+		false,
+		"Print the dashboard URL instead of launching a browser. Useful in headless or CI sessions where no GUI is available.",
+	)
+
+	common.RootCmd.AddCommand(browseCmd)
+}
--- a/cmd/browse/browse.go
+++ b/cmd/browse/browse.go
@@ -14,13 +14,10 @@ import (
 	browsepkg "github.com/bitrise-io/bitrise-build-cache-cli/v2/pkg/browse"
 )
 
-// browseParams collects flag-mapped inputs for the subcommand. Kept in a
-// struct so additional flags can land without expanding the RunE
-// signature. Exported field-by-field for inspection in tests.
-//
-//nolint:gochecknoglobals // mirrors the pattern in cmd/xcode, cmd/gradle, …
+//nolint:gochecknoglobals
 var browseParams struct {
-	printOnly bool
+	workspaceID string
+	printOnly   bool
 }
 
 //nolint:gochecknoglobals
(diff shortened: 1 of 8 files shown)


cmd/browse/browse.go
```diff
@@ -0,0 +1,71 @@
...
+		return nil
+	},
+}
+
+func init() {
+	browseCmd.Flags().StringVar(
+		&browseParams.workspaceID,
+		"workspace",
+		"",
+		"Workspace slug. Falls back to BITRISE_BUILD_CACHE_WORKSPACE_ID env var when empty.",
```
zsolt-marta-bitrise: That will be empty on local, why not the configured per-tool sidecar / auth config?

internal/browse/url.go
```diff
@@ -0,0 +1,52 @@
...
+	// PathEscape per segment + RawPath so a stray `/` or `#` in a workspace/invocation slug can't break the URL.
+	ws := url.PathEscape(p.WorkspaceID)
+	inv := url.PathEscape(p.InvocationID)
+
+	switch {
+	case p.WorkspaceID == "":
+		u.Path = "/build-cache"
+		u.RawPath = ""
+
+	case p.InvocationID != "":
```
zsolt-marta-bitrise: This won't work without a workspace id, we should print an error in that case

internal/browse/url.go
```diff
@@ -0,0 +1,52 @@
...
+		u.Path = "/build-cache/" + p.WorkspaceID + "/invocations/" + p.InvocationID
+		u.RawPath = "/build-cache/" + ws + "/invocations/" + inv
+
+	default:
+		u.Path = "/build-cache/" + p.WorkspaceID + "/invocations"
+		u.RawPath = "/build-cache/" + ws + "/invocations"
+		if p.SourceFilter != "" {
+			q := u.Query()
+			q.Set("source", p.SourceFilter)
```
zsolt-marta-bitrise: What is this param? For local invocations we need the  `ci_provider=unknown` filter. We currently don't have a username filter on the FE/BE

cmd/browse/browse.go
Badlazzor: Done in 9ca1d57 — Browser.Open now falls back to the multiplatform analytics config on disk (where every `activate` flow persists the workspace ID) when both `--workspace` and `BITRISE_BUILD_CACHE_WORKSPACE_ID` are empty. The resolver is injectable via Browser.WorkspaceFromAuth for tests.

internal/browse/url.go
Badlazzor: Done in 9ca1d57 — `BuildURL` now returns a typed `ErrMissingWorkspace` when WorkspaceID is empty. `pkg/browse` still guards above this with `ErrWorkspaceNotConfigured` (the workspace-fallback chain runs first); the new error path is defence-in-depth for any future direct caller of `internal/browse`.

internal/browse/url.go
Badlazzor: Done in 9ca1d57 — the filter for the list view is now `ci_provider=unknown` (param renamed `SourceFilter` → `CIProviderFilter`). The old `source=local` value is gone. When the BE/FE ship a username filter we can stack it on; today this is the closest "my local invocations" proxy.

cmd/browse/browse.go
```diff
@@ -0,0 +1,71 @@
...
+	workspaceID string
+	printOnly   bool
+}
+
+//nolint:gochecknoglobals
+var browseCmd = &cobra.Command{
+	Use:   "browse [invocation-id]",
+	Short: "Open the Bitrise Build Cache dashboard for the configured workspace",
+	Long: `browse opens the user's default browser at the Bitrise Build Cache dashboard, ` +
+		`pre-filtered to the configured workspace (BITRISE_BUILD_CACHE_WORKSPACE_ID) and ` + "`source=local`" + ` invocations. ` +
```
zsolt-marta-bitrise: please update this that we use ci_provider=unknown

cmd/browse/browse.go
zsolt-marta-bitrise: We now have the unified auth config (keychain, oauth), don't we have the ws id stored there too?

pkg/browse/browser.go
```diff
@@ -0,0 +1,112 @@
...
+		InvocationID:     p.InvocationID,
+		CIProviderFilter: ciProvider,
+		BaseURL:          p.BaseURL,
+	})
+	if err != nil {
+		return "", fmt.Errorf("build dashboard URL: %w", err)
+	}
+
+	if b.Logger != nil {
+		b.Logger.Infof("Bitrise Build Cache dashboard: %s", dashboardURL)
```
zsolt-marta-bitrise, line 83: Do we want a json option for automations? Not that important since I don't think it's hard to parse in its form just for the sake of consistency

pkg/browse/browser.go
```diff
@@ -0,0 +1,112 @@
...
+		}
+	}
+
+	return dashboardURL, nil
+}
+
+// defaultWorkspaceFromAuth reads the WorkspaceID from the multiplatform
+// analytics config, which every `activate` flow writes.
+func defaultWorkspaceFromAuth(_ map[string]string) (string, error) {
+	cfg, err := multiplatformconfig.ReadConfig(utils.DefaultOsProxy{}, utils.DefaultDecoderFactory{})
```
zsolt-marta-bitrise: Why not the unified auth package (keychain, oauth)?

cmd/browse/browse.go
Badlazzor: Done in 27b5cb6 — `Long` now spells out the `ci_provider=unknown` filter + the auth-config workspace fallback.

cmd/browse/browse.go
Badlazzor: Done in 87e7370 (earlier) + 27b5cb6 — `defaultWorkspaceFromAuth` calls `configcommon.ResolveAuthConfig` so keychain/OAuth/multiplatform-analytics all feed the same fallback. `--workspace` flag desc updated to match.

pkg/browse/browser.go
Badlazzor: Added in 27b5cb6 — `--json` emits `{url, workspace_id, invocation_id}` on stdout, skips launcher + human log line, so it pipes cleanly into `jq`.

pkg/browse/browser.go
Badlazzor: Done in 87e7370 — `defaultWorkspaceFromAuth` now goes through `configcommon.ResolveAuthConfig` (keychain → OAuth → multiplatform-analytics, single source of truth).

zsolt-marta-bitrise (approved):

PR #374, opened 2026-06-24
Linux 2026 Stack support

benbitrise:
Install the CLI to a location which `ubuntu` can write to. Background: previous versions of the stack ran as `root`, but the 2026 Linux stack runs a `ubuntu`, which does not have permission to write to `/usr/local/bin`


benbitrise:
Closing in favor of https://github.com/bitrise-io/ubuntu-android/pull/720

PR #377, opened 2026-06-25
feat: ACI-5090 shared local invocation log (CLI writer + Xcode hook + schema doc)

Badlazzor:
## Summary

* New `internal/invocations` package: daily NDJSON files under `~/.local/state/bitrise-build-cache/invocations/<YYYY-MM-DD>.ndjson`. Append-only, multi-writer safe (single sub-PIPE_BUF write per record), 30-day retention via `Sweep`.
* `internal/paths.InvocationsDir` / `InvocationsFile(day)` so every writer + reader resolves the path the same way.
* `cmd/xcode.XcodebuildRunner` mirrors the analytics invocation into the local log right before the BE PUT — so a BE outage doesn't lose the local record. Local-write failure logged at warn, never blocks the wrapper.
* `docs/local-invocation-log.md`: schema, concurrency contract, retention, Kotlin reference snippet for gradle-plugins maintainers.

Implements [ACI-5090](https://bitrise.atlassian.net/browse/ACI-5090). Targets the epic branch `aci-4337-cli`.

## Design choices (resolved before code)

* **Format**: NDJSON daily files — atomic POSIX append, easy retention by filename, no schema-migration tax.
* **Path**: `~/.local/state/bitrise-build-cache/invocations/` (consistent with refresh registry + the rest of `StateDir`).
* **Gradle plugin integration**: doc'd schema + reference Kotlin sketch. Each writer owns its append; follow-up patch in `gradle-plugins`.
* **Retention**: 30 days. Trivially enforced over the daily-file layout.

## Out of scope (follow-ups)

* Gradle plugin writer (separate PR in `bitrise-io/gradle-plugins`).
* RN wrapper writer.
* `doctor` / `status` consumers reading from the log.

## Test plan

- [x] `make lint` clean.
- [x] `go test -tags unit -race ./internal/invocations/ ./internal/paths/ ./cmd/xcode/` green.
- [x] Concurrent-append test: 16 writers × 50 records each, no interleave (all lines parseable).
- [x] Daily rotation test: write at `23:59:59Z` then `+2s` → two files.
- [x] Sweep test: old file removed, recent kept.
- [ ] CI green.

[ACI-5090]: https://bitrise.atlassian.net/browse/ACI-5090?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/local-invocation-log.md
```diff
@@ -0,0 +1,104 @@
...
+  2026-06-23.ndjson
+  2026-06-24.ndjson
+  2026-06-25.ndjson   <- today, append-only
+```
+
+One file per UTC date. Files older than 30 days are deleted by the next `Append` from the canonical Go writer (opportunistic sweep gated by a `.last-sweep` marker file, runs at most once every 24h per process).
+
+## Schema
+
+One JSON object per line. UTF-8, LF-terminated. Records ≤ 4 KiB land atomically: POSIX does not formally guarantee regular-file `O_APPEND` atomicity, but Linux and macOS deliver it in practice for writes within PIPE_BUF. Larger records are still written verbatim — the `command` field needs to stay intact for rerun, so writers MUST NOT truncate it — but two concurrent multi-process writes of oversize lines may interleave; readers should treat malformed lines as recoverable parse errors.
```
zsolt-marta-bitrise, line 18: Good idea trying to make it as atomic as possible 👍🏻

internal/invocations/invocations.go
```diff
@@ -0,0 +1,241 @@
...
+type Writer struct {
+	Paths paths.Paths
+	Clock func() time.Time
+}
+
+func NewWriter(p paths.Paths) *Writer {
+	return &Writer{Paths: p}
+}
+
+func (w *Writer) Append(rec Record) error {
```
zsolt-marta-bitrise, line 66: Will users open these logs manually? If so, wdyt printing an info line that we saved the invocation data into the log? If they won't, that case I'd still log as _debug_

internal/invocations/invocations.go
```diff
@@ -0,0 +1,241 @@
...
+		if t.Before(cutoff) {
+			if err := os.Remove(f); err != nil {
+				return removed, fmt.Errorf("remove %s: %w", f, err)
+			}
+
+			removed++
+		}
+	}
+
+	return removed, nil
```
zsolt-marta-bitrise, line 197: Let's debug log the sweep result

zsolt-marta-bitrise (approved):
minor stuff, otherwise LGTM

internal/invocations/invocations.go
Badlazzor: Done in 09a3dea — `Writer.Logger` field; `Append` debug-logs the daily file path after the write succeeds. Wired from `cmd/xcode/xcodebuild.go` so `--debug` surfaces it. Defaulted to debug because the path is stable and users won't typically open the file ad-hoc.

internal/invocations/invocations.go
Badlazzor: Done in 09a3dea — `maybeSweep` debug-logs the count of stale files removed.

zsolt-marta-bitrise (approved):

PR #386, opened 2026-06-29
feat: tier-2 robolectric sandbox probe (-Xlog:class+load), on by default + guarded

zsolt-marta-bitrise:
## What

Adds a **Tier-2** Robolectric probe to the gradle-mirrors init script that observes `android.util.DisplayMetrics` **as Robolectric's sandbox classloader loads it** — the place the SSW-3019 `NoSuchFieldError` actually resolves. **On by default**, sharing the audit's `BITRISE_ROBOLECTRIC_JAR_AUDIT=false` kill-switch (no new env).

## Why Tier-1 wasn't enough

Tier-1 (#385) scans `Test.getClasspath()` — the Gradle **test runtime classpath** — and in production reports a single `android.jar` stub on **both** passing (DDG) and failing (Substack) builds. `android-all-instrumented` is loaded in Robolectric's **sandbox classloader**, invisible to `getClasspath()`. So Tier-1 confirms-by-elimination but can't name the shadow.

## How it works (no agent, no compile)

On a Robolectric test task, the test JVM gets:
```
-Xlog:class+load=info:file=<gradle-home>/bitrise-robolectric-sandbox-probe/<task>-%p.classload.log::filesize=16m,filecount=1
```
The JVM records, for every class it loads in **any** loader (incl. the sandbox loader), its source jar / defining loader. The audit finalizer streams that per-fork file, greps the `android.util.DisplayMetrics` lines, and re-emits them:
```
[robolectric-sandbox-probe] cli=… [class,load] android.util.DisplayMetrics source: <jar | __JVM_DefineClass__ | loader>
```
**Reading it:** `source: __JVM_DefineClass__` = Robolectric instrumented + defined the class in its sandbox (the good path, seen on green DDG); `source: <some android.jar / mockable-android jar>` = a stub supplied DisplayMetrics → that's the shadow behind the `NoSuchFieldError`. This is the discriminator Tier-1 structurally can't produce.

## On by default — why it's safe

It runs fleet-wide on Robolectric test tasks in mirror DCs, so it is hardened to never fail a build (each guard maps to a concrete failure mode):

| Failure mode | Guard |
|---|---|
| JDK 8 aborts on unrecognized `-Xlog` → whole task fails | attach only when the **test JVM is JDK 9+** (read via the task's `javaLauncher` metadata, non-deprecated; daemon-version fallback) |
| `-Xlog` polluting test stdout breaks output-parsing tests | routed to `file=` (never console) |
| disk fill (logs every class load, per fork) | `filesize=16m,filecount=1` (~32 MB/fork) **and** attach only to tasks whose classpath carries `android/util/DisplayMetrics` (skips pure-JVM unit-test tasks) |
| unwritable `GRADLE_USER_HOME` → JVM log-init failure | `mkdirs()` + `isDirectory && canWrite` check before attaching |
| daemon OOM reading a big log | finalizer streams with `useLines` (+ handles rotated `*.classload.log.N`) |
| up-to-date / cache thrash | `jvmArgs` set in `doFirst` → outside the task's input snapshot |
| init-script compile break (the v2.8.10 class) under `allWarningsAsErrors` | non-deprecated APIs only; everything wrapped |

Kill-switch: `BITRISE_ROBOLECTRIC_JAR_AUDIT=false` disables the audit **and** this probe. Any guard failure just skips the probe — it can't fail the build.

**Trade-off (called out):** on by default means a small per-test-JVM tax (`-Xlog:class+load` + bounded file I/O) on every Robolectric task in mirror DCs. The Android-task gate keeps it off the majority of (non-Android) test tasks.

## Validation

Validated end-to-end on the DDG **jarprobe** (app `9e4b8cbb`, sdk=35 Substack-jar run), default-on with no env set:
```
[robolectric-sandbox-probe] cli=… [class,load] android.util.DisplayMetrics source: __JVM_DefineClass__
```
Confirms: opt-out wiring fires with no env, `jvmArgs` set in `doFirst` reaches the fork, JDK-9+/Android-task gates pass, init script compiles on real Gradle, build SUCCESSFUL. (DDG is green / no shadow, so this proves the mechanism + safety; a failing Substack-class build is where it captures the shadow `source:`.)

`make lint` clean; mirror unit tests updated + passing.

### DIFF
--- a/internal/config/gradle/mirrors/asset/gradle-mirrors.init.gradle.kts.gotemplate
+++ b/internal/config/gradle/mirrors/asset/gradle-mirrors.init.gradle.kts.gotemplate
@@ -85,6 +85,9 @@ class InternalRepositoryPlugin : Plugin<Gradle> {
             {{- range .Mirrors }}{{ if .UseAsRobolectricRepo }}
             // Root-only: wire the Robolectric android-all redirect + jar audit across every Test task in the build, registered and logged once (avoids per-subproject flooding).
             if (getParent() == null) {
+                // Tier-2 sandbox probe (opt-in: BITRISE_ROBOLECTRIC_SANDBOX_PROBE=true). The Tier-1 classpath probe only sees the Gradle test runtime classpath; android.util.DisplayMetrics is actually loaded inside Robolectric's sandbox classloader. When enabled we add -Xlog:class+load to each Test JVM so the VM itself records the source jar / defining loader of every class it loads (to a per-fork file); the audit finalizer then greps the android.util.DisplayMetrics lines out and re-emits them — naming the loader/jar behind the NoSuchFieldError. No agent, no compile; never fails the build.
+                val sandboxProbe{{ .ID }} = System.getenv("BITRISE_ROBOLECTRIC_SANDBOX_PROBE") == "true"
+                val sandboxLogDir{{ .ID }} = java.io.File(System.getenv("GRADLE_USER_HOME") ?: (System.getProperty("user.home") + "/.gradle"), "bitrise-robolectric-sandbox-probe")
                 // Finalizer task (runs even when tests FAIL) logs the resolved android-all jar's sha1 once; set BITRISE_ROBOLECTRIC_JAR_AUDIT=false to opt out. Wrapped so it never fails the build.
                 val auditTask{{ .ID }} = if (System.getenv("BITRISE_ROBOLECTRIC_JAR_AUDIT") != "false") {
                     tasks.register("bitriseRobolectricJarAudit{{ .ID }}") {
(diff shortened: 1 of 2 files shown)


internal/config/gradle/mirrors/asset/gradle-mirrors.init.gradle.kts.gotemplate
```diff
@@ -150,6 +165,13 @@ class InternalRepositoryPlugin : Plugin<Gradle> {
                                         .map { "${it.getName()}(${it.length()})" }
                                     if (dmSources.isNotEmpty()) {
                                         getLogger().lifecycle("[robolectric-classpath-probe] cli={{ $.CLIVersion }} ${getPath()} android/util/DisplayMetrics from: ${dmSources.joinToString(", ")}")
+                                        // Tier-2: this is an Android/Robolectric test task. Attach -Xlog:class+load so the test JVM records where DisplayMetrics is actually defined (incl. Robolectric's sandbox loader). Guards: JDK 9+ only (the flag aborts an unrecognized-option JDK 8 JVM), writable dir, size-capped file (never stdout). jvmArgs in doFirst is read at fork time (after this action) yet stays out of up-to-date inputs.
```
zsolt-marta-bitrise, line 168: Long, but this is really complex unorthodox code to squeeze out some info about what's happening. I'll clean everything up when we have some kind of resolution

Badlazzor (approved):

PR #390, opened 2026-06-30
feat: SSW-3019 capture native-runtime failure cause in gradle-mirrors init script

zsolt-marta-bitrise:
## Why
SSW-3019 — Robolectric `NoSuchFieldError@DeviceConfig` mass failures across ≥7 customers. Investigation (see SSW-3019 notes) established:
- The `NoSuchFieldError` is a **downstream cascade**. Bucket A starts with `NPE@DefaultNativeRuntimeLoader:201`, which in robolectric 4.16 is `String.join(",", getCoreClassNatives())` over a **null** — i.e. an earlier native-runtime/ICU/`.so` load failed silently and was swallowed.
- It's **intermittent and self-heals**, and the KV cache archive hash is identical on clean and failing runs — so it is **not** a statically poisoned cache artifact. Every android-all jar we can currently measure is healthy.
- We have **no SSH access** to a live failing VM, so the init script is our only instrument.

## What
Extend the always-on (opt-out via `BITRISE_ROBOLECTRIC_JAR_AUDIT=false`) probe so the **next** classifier-flagged failure self-captures the real cause. The audit already runs on failing builds (`finalizedBy` runs on task failure), so this captures failure-time state without a repro:

- **Audit ICU entry** — report `icu/icudt*.dat` presence+size per android-all jar (the resource the native runtime reads). A DisplayMetrics-present jar with a MISSING/short ICU entry is the bad-bytes smoking gun the DisplayMetrics check alone misses. Central-directory scan only → cheap on every build.
- **`-Xlog:library`** added to the existing per-fork `-Xlog` file; finalizer re-emits native-runtime load lines as `[robolectric-nativeruntime-probe]`, filtered to `androidruntime/nativeruntime/icudt/Failed` so it can't echo every `.so`. Tells us whether/where `libandroidruntime.so` loaded.
- **`testLogging FULL` + showCauses on the `failed` event only** — surfaces the swallowed upstream cause; **zero output on green builds**, additive (no event-set clobber).
- **`robolectric.logging=stdout`** gated behind opt-in **`BITRISE_ROBOLECTRIC_VERBOSE_LOGGING=true`** (default OFF) — it logs on every test in every fork and would flood the fleet, so it's available only for a targeted repro.

## Flood safety
Only the opt-in stdout logging is flood-prone, and it's off by default. The rest add output only on failing builds (`testLogging` failed-only) or a handful of filtered lines (`-Xlog` re-emit + audit). Verifying on the `e2e-mavencentral-mirror-duckduck` e2e that `logs.txt` size is unchanged vs the previous run (this PR is draft pending that check).

## Verified
- `make lint` clean, `make test-unit` green.
- Rendered init script compiles clean under `org.gradle.kotlin.dsl.allWarningsAsErrors=true` (Gradle 9.3 / JDK 17) — the v2.8.10-regression guard.


internal/config/gradle/mirrors/asset/gradle-mirrors.init.gradle.kts.gotemplate
```diff
@@ -118,22 +118,36 @@ class InternalRepositoryPlugin : Plugin<Gradle> {
                                         }
                                         val sha1 = md.digest().joinToString("") { String.format("%02x", it.toInt() and 0xFF) }
                                         // Structural check: an intact-but-wrong jar (or a truncated one) shows up as MISSING/zip-error or an off DisplayMetrics size, distinguishing it from a correct jar that fails for classpath/version reasons.
+                                        // Also report the ICU data entry: DefaultNativeRuntimeLoader reads icu/icudt*.dat from android-all, and a bucket-A NPE@DefaultNativeRuntimeLoader is an ICU/native read that returned null. A DisplayMetrics-present jar with a MISSING/short icu entry is the smoking gun the displayMetrics check alone misses. Central-directory scan only (no entry-content read), so it stays cheap on every build.
```
zsolt-marta-bitrise, line 121: Long, but temporary and unorthodox+complex

Badlazzor (approved):

PR #394, opened 2026-07-02
ACI-4264: username override for local invocations (env + keychain + wizard)

Badlazzor:
## Summary
- `keychain.Credentials.Username` — optional persisted display name.
- `configcommon.ResolveUsername(envs)` — precedence `BITRISEIO_CACHE_USERNAME` env → keychain → `os/user.Current().Username`.
- `bitrise-build-cache auth set --username <name>` — non-interactive persist.
- `activate --interactive` — new wizard prompt "Display name for this machine's local invocations". Defaults to OS username, empty allowed. Updates keychain when the user modifies the stored value.

## Out of scope
- Consumers that stamp invocations with the resolved username (tracked as ACI-5046).
- `invocations list` CLI filter + BE-side filter.

## Test plan
- [x] `make lint` clean.
- [x] `go test -tags unit -race ./internal/config/common/... ./internal/auth/keychain/... ./cmd/auth/... ./cmd/common/...` green.
- [x] Unit tests for `resolveUsername` cover env-wins / keychain-wins / OS fallback / whitespace-empty env skip / all-empty.
- [ ] Manual: run `bitrise-build-cache auth set --token X --workspace-id Y --username alice`, then `bitrise-build-cache auth status` shows the display name.
- [ ] Manual: `activate --interactive` → keychain-only path → change display name → confirm keychain updated.

### DIFF
--- a/internal/config/common/auth.go
+++ b/internal/config/common/auth.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"os/user"
 	"strings"
 
 	"github.com/bitrise-io/bitrise-build-cache-cli/v2/internal/auth/keychain"
@@ -14,6 +15,7 @@ const (
 	EnvAuthToken   = "BITRISE_BUILD_CACHE_AUTH_TOKEN"   //nolint:gosec // env-var key, not a credential
 	EnvWorkspaceID = "BITRISE_BUILD_CACHE_WORKSPACE_ID" //nolint:gosec // env-var key, not a credential
 	EnvJWT         = "BITRISEIO_BITRISE_SERVICES_ACCESS_TOKEN"
+	EnvUsername    = "BITRISE_BUILD_CACHE_USERNAME"
 )
 
 var (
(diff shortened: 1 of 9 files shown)


internal/config/common/auth.go
```diff
@@ -14,6 +15,7 @@ const (
 	EnvAuthToken   = "BITRISE_BUILD_CACHE_AUTH_TOKEN"   //nolint:gosec // env-var key, not a credential
 	EnvWorkspaceID = "BITRISE_BUILD_CACHE_WORKSPACE_ID" //nolint:gosec // env-var key, not a credential
 	EnvJWT         = "BITRISEIO_BITRISE_SERVICES_ACCESS_TOKEN"
+	EnvUsername    = "BITRISEIO_CACHE_USERNAME"
```
zsolt-marta-bitrise: nit: I'd go with `BITRISE_BUILD_CACHE_` prefix as with the token/wsid. the services access token (jwt) is for CI

zsolt-marta-bitrise (approved):

internal/config/common/auth.go
Badlazzor: Done in ab069df — renamed to `BITRISE_BUILD_CACHE_USERNAME` for prefix consistency with token/workspace-id.

PR #397, opened 2026-07-02
ACI-5127: daemon info reports per-service status

Badlazzor:
## Summary
- `daemon info` now reports `running` / `stopped` / `not configured` per service, computed via `stat` on the socket path + a 500 ms unix-socket dial (proper ccache health-check for the storage helper so the probe doesn't pollute its log).
- Text output: socket lines unchanged, followed by a `xcelerate-proxy status: ...` / `ccache-helper status: ...` block.
- JSON output: existing string keys (`xcelerateProxy`, `ccacheHelper`) unchanged for backward compat; adds sibling `xcelerateProxyStatus` / `ccacheHelperStatus`.
- e2e assertion (`e2e-daemon-install-macos`) updated: anchors socket-capture regex, checks both new status lines + both new JSON fields.

## Test plan
- [x] `make lint` clean.
- [x] `go test -tags unit -race ./cmd/daemon/...` green — `TestProbeSocket_*` covers missing file / stale file / listening socket / accept-and-hang; `TestProbeCcacheSocket_*` smoke-tests the health-check path.
- [ ] CI e2e-daemon-install-macos passes with the updated assertions.

## Out of scope
- Underlying launchd/systemd unit status (e.g. `launchctl print`) — the socket-dial probe is what actually matters for downstream tools; unit-loaded-but-crashed will still show as `stopped`.
- Follow-up: on next JSON-schema bump, emit `""` for `xcelerateProxy` / `ccacheHelper` when the service is not configured (current placeholder string kept for backward compat; `*Status == "not configured"` is the cleaner signal now).

### DIFF
--- a/cmd/daemon/info.go
+++ b/cmd/daemon/info.go
@@ -1,13 +1,20 @@
 package daemon
 
 import (
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"io/fs"
+	"net"
+	"os"
+	"time"
 
+	"github.com/bitrise-io/go-utils/v2/log"
 	"github.com/spf13/cobra"
 
+	"github.com/bitrise-io/bitrise-build-cache-cli/v2/cmd/common"
+	"github.com/bitrise-io/bitrise-build-cache-cli/v2/internal/ccache"
 	ccacheconfig "github.com/bitrise-io/bitrise-build-cache-cli/v2/internal/config/ccache"
 	xcelerateconfig "github.com/bitrise-io/bitrise-build-cache-cli/v2/internal/config/xcelerate"
 	"github.com/bitrise-io/bitrise-build-cache-cli/v2/internal/utils"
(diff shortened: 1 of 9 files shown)


cmd/daemon/info.go
```diff
@@ -47,38 +73,88 @@ var infoCmd = &cobra.Command{
...
+	}
+}
+
+func probeSocket(path string) string {
+	if _, err := os.Stat(path); err != nil {
+		if !errors.Is(err, fs.ErrNotExist) {
+			debugLogger().Debugf("probeSocket: stat %s failed: %v", path, err)
+		}
+
+		return statusStopped
```
zsolt-marta-bitrise, line 117: Can these mean they are also stuck, i.e. they need to be stopped before restarting as part of a fix?

Badlazzor: Good catch. Added a third status `stuck` (socket present but Dial / SendHealthCheck fails — stale socket from crashed helper or zombie holding the socket). Message points at `doctor --fix` first, `daemon restart` as fallback. Doctor's ccache check switched from `RemoveFileFixer` to `DaemonRestartFixer` — with the new KeepAlive={SuccessfulExit=false} in this same PR, launchd would not respawn a clean-exited helper, so removing the socket without restart left it down. Fixed in b7f0f3a.

zsolt-marta-bitrise (approved):

PR #399, opened 2026-07-02
feat: ACI-5041 + ACI-5042 xcode-app enable/disable subcommands

Badlazzor:
## Summary

- `bitrise-build-cache xcode-app enable` / `disable` — Xcode.app (GUI) builds pick up the compile cache without the `xcodebuild` wrapper. Targets #1 iOS-customer ask (Foreflight / Redfin / Depop / Expedia).
- `enable` writes override `xcconfig` under `~/.bitrise-xcelerate/`, `launchctl setenv XCODE_XCCONFIG_FILE`, registers a login LaunchAgent that re-applies `setenv`, installs + starts `xcelerate-proxy` daemon. Detects running Xcode → relaunch nudge.
- ACI-5042 folded in: pre-existing `XCODE_XCCONFIG_FILE` gets `#include`-chained at top of override + captured in state file for restore on `disable`.

Re-land of closed [PR #366](https://github.com/bitrise-io/bitrise-build-cache-cli/pull/366). Rebased onto current main via `aci-4337-cli-xcode` epic base. 5 commits preserved (feat + review-fix + 3 refactor/fixups) — see commit log.

## Mechanism

Per [E1 spike doc](docs/xcode-companion-app-spike-2026-05-22.md): override slot wins over per-project/per-target xcconfigs, sits below `xcodebuild` CLI args, so wrapper flow unaffected. xcconfig writes:

- `COMPILATION_CACHE_REMOTE_SERVICE_PATH` → xcelerate-proxy unix socket (from `~/.bitrise-xcelerate/config.json`)
- All keys from `internal/xcelerate/xcodeargs.CacheArgs`

## Layout

- `cmd/xcode_app/{xcode_app,enable,disable}.go` — cobra wrapper
- `pkg/xcode_app/activator.go` — public `Activator`, injectable collaborators
- `internal/xcode_app/`: `paths.go`, `xcconfig.go`, `state.go`, `launchctl.go`, `launch_agent.go`, `xcode_running.go`

## Sentinels

- `xcode_app.ErrUnsupportedPlatform` — darwin-only.
- `xcode_app.ErrXcelerateNotConfigured` — requires `~/.bitrise-xcelerate/config.json`; hint points at `activate xcode`.

## Test plan

- [ ] `make check` local — done pre-push.
- [ ] macOS manual: `activate xcode` → `xcode-app enable` → verify xcconfig written, `launchctl getenv XCODE_XCCONFIG_FILE` set, LaunchAgent bootstrapped, proxy running.
- [ ] Relaunch Xcode, build small project, proxy log shows hits.
- [ ] `disable` reverses everything.
- [ ] `#include` chain: pre-existing `XCODE_XCCONFIG_FILE` → `enable` chains → `disable` restores.

## Out of scope

- E4 docs (ACI-5043) — separate PR stacked on this.
- F1/F2 analytics (ACI-5044/ACI-5045).
- Xcode-beta / multi-Xcode coverage — `pgrep -x Xcode` matches GA only.


internal/xcode_app/launchctl.go
```diff
@@ -0,0 +1,112 @@
...
+	if c.Bin != "" {
+		return c.Bin
+	}
+
+	return LaunchctlBin
+}
+
+// Setenv lasts only until logout — pair with the LaunchAgent to survive logout.
+func (c LaunchctlClient) Setenv(ctx context.Context, key, value string) error {
+	_, stderr, code, err := c.runner().Run(ctx, c.bin(), "setenv", key, value)
```
zsolt-marta-bitrise, line 42: can we consolidate with the envexporter package, or is this very specific to local xcode?

internal/xcode_app/launch_agent.go
```diff
@@ -0,0 +1,93 @@
...
+	"io/fs"
+	"path/filepath"
+	"strings"
+
+	"github.com/bitrise-io/bitrise-build-cache-cli/v2/internal/utils"
+)
+
+const SetenvAgentLabel = "io.bitrise.build-cache.xcode-app-setenv"
+
+const SetenvAgentPlistRelative = "Library/LaunchAgents/" + SetenvAgentLabel + ".plist"
```
zsolt-marta-bitrise: Please add to the paths pacakge so we can track which files are being created by the CLI

pkg/xcode_app/activator.go
```diff
@@ -0,0 +1,315 @@
...
+	return nil
+}
+
+// resolvePreviousOverride: on self-loop (env == ownPath) fall back to on-disk state so the original prior override survives repeat Enables.
+func resolvePreviousOverride(envValue, statePath, ownPath string) string {
+	if envValue != ownPath {
+		return envValue
+	}
+
+	existing, found, err := xa.LoadState(statePath)
```
zsolt-marta-bitrise, line 253: Let's log this so that the user is aware that it's not lost

pkg/xcode_app/activator.go
```diff
@@ -0,0 +1,315 @@
...
+
+	if err := osProxy.MkdirAll(xcelerate.DirPath(osProxy), 0o755); err != nil {
+		return EnableResult{}, fmt.Errorf("mkdir xcelerate dir: %w", err)
+	}
+
+	if err := osProxy.WriteFile(xcconfigPath, []byte(body), 0o644); err != nil { //nolint:gosec // xcconfig must be readable by Xcode
+		return EnableResult{}, fmt.Errorf("write %s: %w", xcconfigPath, err)
+	}
+
+	if err := xa.SaveState(statePath, xa.State{PreviousXCConfigPath: previous}); err != nil {
```
zsolt-marta-bitrise, line 96: Please add debug logs for steps along the way, info for the major milestones. Applies to the other commands too

pkg/xcode_app/activator.go
```diff
@@ -0,0 +1,315 @@
...
+		PreviousXCConfigPath: state.PreviousXCConfigPath,
+	}
+
+	if state.PreviousXCConfigPath != "" {
+		// Restore the user's prior override so their tooling keeps working.
+		if err := a.Launchctl.Setenv(ctx, xa.XCConfigEnvVar, state.PreviousXCConfigPath); err != nil {
+			return result, fmt.Errorf("restore previous XCODE_XCCONFIG_FILE: %w", err)
+		}
+
+		result.RestoredXCConfigPath = state.PreviousXCConfigPath
```
zsolt-marta-bitrise, line 189: Log this too so that the user is aware

internal/xcode_app/xcconfig.go
```diff
@@ -0,0 +1,52 @@
...
+
+	var b strings.Builder
+
+	b.WriteString("// Bitrise Build Cache — Xcode.app override\n")
+	b.WriteString("// Written by `bitrise-build-cache xcode-app enable`.\n")
+	b.WriteString("// Removed by `bitrise-build-cache xcode-app disable`.\n")
+	b.WriteString("// Do not edit by hand.\n\n")
+
+	if previousIncludePath != "" {
+		// Chain the prior override so we don't clobber the user's setup. Quoted form supports paths with spaces.
```
zsolt-marta-bitrise, line 31: Nice thinking that we just add overrides

pkg/xcode_app/activator.go
```diff
@@ -0,0 +1,315 @@
...
+	if cfg.ProxySocketPath == "" {
+		return EnableResult{}, fmt.Errorf("xcelerate config has empty proxy socket path — re-run `activate xcode`")
+	}
+
+	xcconfigPath := xcelerate.XcodeAppOverrideXCConfigFile(osProxy)
+	statePath := xcelerate.XcodeAppStateFile(osProxy)
+
+	previous := resolvePreviousOverride(a.Envs[xa.XCConfigEnvVar], statePath, xcconfigPath)
+
+	body, err := xa.RenderOverride(cfg.ProxySocketPath, previous)
```
zsolt-marta-bitrise, line 82: 1. Is this socket path the same across restarts? so that when launchctl launches the xcconfig override, the xcconfig file will still be valid?

2. The daemon is restarted on login, right? Why not call setenv from there instead of installing a separate launchctl agent?

internal/xcode_app/launch_agent.go
Badlazzor: Done in 24f70de — added `XcodeAppSetenvAgentLabel` + `XcodeAppSetenvAgentPlistFile()` to `internal/paths`, callers now go through it.

pkg/xcode_app/activator.go
Badlazzor: Added in ebbbc2b — Enable() now logs `Preserving prior XCODE_XCCONFIG_FILE override: <path>` at Info when a prior override is chained, plus a Debug at the self-loop recovery path in `resolvePreviousOverride`.

pkg/xcode_app/activator.go
Badlazzor: Added in ebbbc2b — Enable() and Disable() now emit `TInfof` on entry, `Infof` on the major milestones (LaunchAgent bootstrap, daemon start, prior-override preservation, prior-override restore), and `Debugf` on each step (config read, xcconfig write, state save, launchctl setenv/unsetenv, plist write, bootout, override remove).

pkg/xcode_app/activator.go
Badlazzor: Added in ebbbc2b — Disable() now logs `Restored prior XCODE_XCCONFIG_FILE: <path>` at Info immediately after the launchctl setenv succeeds.

internal/xcode_app/launchctl.go
Badlazzor: Different surfaces — no overlap to consolidate. `envexport.EnvExporter` targets shell/CI env: `os.Setenv` (current process), `envman add` (Bitrise CI), `GITHUB_ENV` file (GH Actions), and `~/.bashrc`/`~/.zshrc` blocks. `launchctl setenv` targets macOS launchd's per-user GUI session, which is the *only* channel Xcode.app (launched from Finder/Dock/Spotlight) reads at process start — envman/GITHUB_ENV/shell RC files never reach a GUI-launched process. So this is Xcode.app-specific and stays in the launchctl client.

pkg/xcode_app/activator.go
Badlazzor: **1. Socket path stability:** yes. `cfg.ProxySocketPath` is written to `~/.bitrise-xcelerate/config.json` at `activate xcode` time, and both `xcode-app enable` (rendering the xcconfig) and the daemon (binding the listener) read that same field, so they can't drift. Default resolves via `os.TempDir()` — on macOS that's the per-user `/var/folders/…/T/xcelerate-proxy.sock`, stable across reboot/logout. Only way to invalidate: user re-runs `activate xcode` with a new `--proxy-socket-path` / `BITRISE_XCELERATE_PROXY_SOCKET_PATH` — then re-run `xcode-app enable` to rewrite the xcconfig.

**2. Daemon vs separate LaunchAgent:** deliberately separate:
- **Independent lifecycle.** `daemon down` for maintenance shouldn't unset `XCODE_XCCONFIG_FILE`; `xcode-app disable` shouldn't kill a daemon the `xcodebuild` wrapper flow still needs. Two plists = two orthogonal switches.
- **Scope.** `daemon.Service` for xcelerate-proxy is shared with ccache-helper etc. Attaching an Xcode.app-specific `launchctl setenv` to daemon startup would leak GUI-only concern into the generic daemon.
- **Plist purity.** Daemon plist uses verbatim `ProgramArguments` (no `sh -c …`) to avoid quoting hazards on paths with spaces. Making it `setenv-then-exec` would need a wrapper script or a Go-side pre-startup call, both of which regress that guarantee.
- **Timing.** LaunchAgent with `RunAtLoad=true` fires early at login with no server startup on the critical path — Xcode.app launched right after login sees the env. Chaining it behind daemon boot adds an avoidable window.

zsolt-marta-bitrise (approved):

PR #412, opened 2026-07-06
fix: ACI-5174 wizard writes gradle + bazel sidecars

Badlazzor:
## Summary

`activate --interactive` was leaving `~/.bitrise/cache/gradle/config.json` and `~/.bitrise/cache/bazel/config.json` unwritten. `refresh.Notify`'s drift-nudge never fires for those tools because the file it inspects didn't exist for wizard-only users.

## Changes

- `cmd/common/activate_interactive.go`
  - `runInteractiveGradle` now writes the gradle sidecar after activation succeeds — same params + best-effort semantics as `cmd/gradle/activate_gradle.go:80`.
  - `runInteractiveBazel` now writes the bazel sidecar after `.bazelrc` is written — mirrors `cmd/bazel/activate_bazel.go:94`.

## Follow-up (not in scope)

Two callers per tool now use an identical WriteSidecar block. Worth extracting `WriteActivatedSidecar` helpers on `gradleconfig` / `bazelconfig` so the shape can't drift when a new field is added. Small refactor, separate PR.

## Test plan

- [x] `go test -tags unit -race ./cmd/common/` passes.
- [x] `make lint-fix` clean.
- [x] End-to-end proxied via direct-activate path — identical `WriteSidecar` call is already exercised by the existing golden gradle/bazel unit tests. Wizard-specific integration tests would need HOME injection infrastructure that doesn't exist yet (out of scope).

Ticket: [ACI-5174](https://bitrise.atlassian.net/browse/ACI-5174)

[ACI-5174]: https://bitrise.atlassian.net/browse/ACI-5174?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cmd/common/activate_interactive.go
+++ b/cmd/common/activate_interactive.go
@@ -17,6 +17,7 @@ import (
 	bazelconfig "github.com/bitrise-io/bitrise-build-cache-cli/v3/internal/config/bazel"
 	gradleconfig "github.com/bitrise-io/bitrise-build-cache-cli/v3/internal/config/gradle"
 	"github.com/bitrise-io/bitrise-build-cache-cli/v3/internal/config/xcelerate"
+	"github.com/bitrise-io/bitrise-build-cache-cli/v3/internal/paths"
 	"github.com/bitrise-io/bitrise-build-cache-cli/v3/internal/utils"
 	ccachepkg "github.com/bitrise-io/bitrise-build-cache-cli/v3/pkg/ccache"
 )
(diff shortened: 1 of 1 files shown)


cmd/common/activate_interactive.go
```diff
@@ -114,6 +114,18 @@ func runInteractiveGradle(logger log.Logger, envs map[string]string, pushEnabled
 		return fmt.Errorf("activate plugins for Gradle: %w", err)
 	}
 
+	if home, homeErr := os.UserHomeDir(); homeErr == nil {
+		initFile := filepath.Join(gradleHome, "init.d", "bitrise-build-cache.init.gradle.kts")
```
zsolt-marta-bitrise: please use the common paths package for all file operations so we can track the files written by the CLI

Badlazzor: Done in f4d37b4 — uses `paths.FromHome(home).GradleInitScriptFile()`.

zsolt-marta-bitrise (approved):

PR #416, opened 2026-07-07
fix: ACI-5176 deduplicate "Saved to keychain" log line

Badlazzor:
## Summary

The interactive wizard printed "Saved auth credentials to the OS keychain" twice for any run that activated both Xcode and ccache — each `Activate` path logged independently. Route both through a shared `LogKeychainSaved` guarded by `sync.Once`.

## Changes

- `internal/config/common/keychain_save_log.go` (new) — `LogKeychainSaved(logger)`, `sync.Once`-guarded.
- `internal/config/xcelerate/activate.go` — use `configcommon.LogKeychainSaved`.
- `pkg/ccache/activator.go` — same.

The keychain writes themselves stay independent + idempotent; only the log is collapsed.

## Test plan

- [x] `go test -tags unit -race ./internal/config/common/ ./internal/config/xcelerate/` passes.
- [x] `make lint-fix` clean.
- [x] `pkg/ccache/` has an unrelated env-dependent flake on `TestActivator_Activate/returns_error_when_auth_config_is_missing` that pre-dates this branch (fails on main too when `BITRISE_BUILD_CACHE_AUTH_TOKEN` is set in the shell). Not this PR's regression.

Ticket: [ACI-5176](https://bitrise.atlassian.net/browse/ACI-5176)

[ACI-5176]: https://bitrise.atlassian.net/browse/ACI-5176?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/config/xcelerate/activate.go
```diff
@@ -101,7 +101,7 @@ func Activate(
 	if keychainErr != nil {
 		mpCfg.AuthConfig = config.AuthConfig
 	} else {
-		logger.Infof("Saved auth credentials to the OS keychain")
+		configcommon.LogKeychainSaved(logger)
```
zsolt-marta-bitrise: Can't we just make keychain saving idempotent (check if it's already saved)? I don't want to introduce this pattern of mutexing just for the sake of logging, we can get a ton of these

Badlazzor: Good call — pushed d6d1f43 that swaps the sync.Once for a Keychain.SaveIfChanged that Loads + compares first and only writes on change. Both callers log only when a write actually happened. Dropped the sync.Once helper file entirely.

zsolt-marta-bitrise (approved):

PR #424, opened 2026-07-13
[NOJIRA] Atomic CLI copy on Xcode activation to avoid corrupted binary

zsolt-marta-bitrise:
[NOJIRA] — reported via Slack (Stripe), no tracking ticket.

## Summary

On a repeated `activate-build-cache-for-xcode`, builds failed with a corrupted CLI left in `~/.bitrise-xcelerate/bin/bitrise-build-cache-cli`:

```
Terminating already running CLI (pid: 9636)
Error: failed to copy xcelerate cli to ~/.bitrise-xcelerate/bin: failed to copy executable: write .../bitrise-build-cache-cli: input/output error
```

`copyCLIToXcelerateBinDir` overwrote the target **in place** (`O_TRUNC` + `io.Copy`) while a previous CLI (the xcelerate proxy) may still be executing it. On darwin, writing into a running executable's text segment yields `ETXTBSY` / mid-write `input/output error`, and because `O_TRUNC` already emptied the file, a **truncated binary is left behind** — breaking the next build. Compounding it, the terminate step sent SIGTERM but did not wait for the old process to exit.

## Changes

- Write the CLI to a temp file in the bin dir, then `os.Rename` it over the target (the repo's existing atomic-write idiom, as in `sidecar.go` / `versioncheck/state.go`). Rename swaps the directory entry, so a failed copy never touches the installed binary and a lingering old process keeps its inode intact.
- After terminate/kill, wait (bounded) for the old CLI process to actually exit before proceeding.
- Add regression tests around the extracted `writeExecutableAtomically` helper.
- **CI (temporary):** comment out the two `rde-*-local-dx-smoke` workflows from the `features-e2e` pipeline while RDE provisioning is failing. Revert once it recovers.

## Testing

- `go test -tags unit -race ./internal/config/xcelerate/` — pass.
- `make lint` — 0 issues.
- Verified the regression test has teeth: swapping the old in-place `O_TRUNC` logic back in makes `FailedCopyLeavesTargetIntact` fail, leaving the target as `"partial"` (the reported corruption); the atomic version keeps it intact.
- Full `go test -tags unit ./...` is green except a pre-existing, environment-dependent failure in `pkg/ccache` (`TestActivator_Activate/returns_error_when_auth_config_is_missing` — auth resolves from the local OS keychain); it fails identically on `origin/main` and is unrelated to this change.

## Notes for reviewers

- The exact production `input/output error` is a race + filesystem-specific effect, so it isn't reproduced deterministically in a test. The tests instead assert the invariant the fix guarantees: a failed/interrupted copy never leaves a corrupted binary at the target.
- Stripe's `run_if` workaround (skip the second activation) can be dropped once a release with this fix ships, but it's harmless to keep.
- Follow-up (not in this PR): `internal/daemon/binary.go` `CopyCLIToStableBin` has the identical in-place-truncate pattern copying to `~/.bitrise/bin` and would benefit from the same fix.

### DIFF
--- a/bitrise.yml
+++ b/bitrise.yml
@@ -2041,8 +2041,9 @@ pipelines:
       react-seek-android-e2e: {}
       react-seek-ios-e2e: {}
       e2e-mavencentral-mirror-duckduck: {}
-      rde-mac-local-dx-smoke: {}
-      rde-linux-local-dx-smoke: {}
+      # TEMP: RDE smoke tests disabled while the RDE provisioning is failing.
+      # rde-mac-local-dx-smoke: {}
+      # rde-linux-local-dx-smoke: {}
     triggers:
       pull_request:
         - source_branch: "*"
--- a/internal/config/xcelerate/activate.go
+++ b/internal/config/xcelerate/activate.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/bitrise-io/go-utils/v2/log"
 	"github.com/shirou/gopsutil/v4/process"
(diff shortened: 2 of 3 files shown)


zsolt-marta-bitrise:
[//]: # ([skip ci])



## Summary
Fixes atomic CLI copy on Xcode activation to prevent corrupted binaries. Replaces in-place file truncation with atomic writes (temp file + rename), adds process exit waiting before binary replacement, and includes regression tests. Temporarily disables RDE smoke tests in CI while provisioning is failing.


## Walkthrough
| File | Summary |
|------|---------|
| internal/config/xcelerate/activate.go | Implemented atomic CLI binary copying via temp file + rename, added process exit waiting (5s bounded) after termination, and extracted writeExecutableAtomically helper to prevent corrupted binaries on failed/interrupted copies |
| internal/config/xcelerate/activate_internal_test.go | Added regression tests for atomic CLI copy, verifying failed copies don't corrupt the target binary |
| bitrise.yml | Temporarily commented out rde-mac-local-dx-smoke and rde-linux-local-dx-smoke workflows in features-e2e pipeline due to RDE provisioning failures |

Badlazzor (approved):

PR #423, opened 2026-07-10
feat: ACI-5045 + ACI-5205 xcactivitylog enrichment re-PUT + F1 inactivity-timer flush

Badlazzor:
## Summary

Combines the F1 inactivity-timer slim emit (ACI-5205), the F2 xcactivitylog enrichment re-PUT (ACI-5045), and the xcelerate wrapper prefix-map + managed DerivedData injection (ACI-5129). Together they close the wrapper-path + Xcode.app-path gap in dashboard reporting and stabilize the compilation-cache keys across absolute-path variants.

> **Scope reduction — xcode-app removed (2026-07-22, `9d1d537`).** Xcode 26 IDE remote CAS via `XCODE_XCCONFIG_FILE` is a documented dead-end (Xcode's build system drops `COMPILATION_CACHE_REMOTE_SERVICE_PATH` before writing `.cas-config`; see [`docs/xcode-app-ide-remote-cas-findings-2026-07-21.md`](../blob/aci-5045-xcactivitylog-enrichment/docs/xcode-app-ide-remote-cas-findings-2026-07-21.md)). The `xcode-app` command surface (`cmd/xcode_app`, `pkg/xcode_app`, `internal/xcode_app`, the doctor xcconfig check, `docs/xcode-app.md`) is removed in this PR. Xcode.app story reduces to `xcodebuild` invocations through the wrapper installed by `activate xcode`. The F1 slim + F2 enrichment paths, the wrapper prefix-map injection, and the proxy daemon all stay.

### ACI-5205 — F1 inactivity-timer flush (was #425)

- `Proxy` grows an `InactivityTimeout` (default 10s) + `time.Timer` field
- gRPC `UnaryInterceptor` calls `touchSession()` after every handler → bumps `lastActivity` + arms timer on first touch
- Timer callback re-schedules if activity refreshed since scheduled fire; otherwise emits slim invocation via existing `emitCurrentSessionLocked`
- Pointer comparison on captured `SessionMeta` target guards against stale callbacks after SetSession swap
- Xcode.app path unaffected (no SetSession → no session → timer never arms)
- Duration uses `meta.EndTime` (last-activity wall-clock) instead of `time.Since(StartTime)` so inactivity-flushed rows report the accurate elapsed span

### ACI-5045 — xcactivitylog enrichment re-PUT

- xcelerate-proxy daemon polls `~/Library/Developer/Xcode/DerivedData/*/Logs/Build/LogStoreManifest.plist` every 5s
[...]

PR #425, opened 2026-07-13
feat: ACI-5205 F1 slim emit inactivity-timer flush

Badlazzor:
## Summary

- `Proxy` grows an `InactivityTimeout` (default 30s) + `time.Timer` field
- gRPC `UnaryInterceptor` calls `touchSession()` after every handler → bumps `lastActivity` + arms timer on first touch
- Timer callback re-schedules if activity refreshed since scheduled fire; otherwise emits slim invocation via existing `emitCurrentSessionLocked`
- Pointer comparison on captured `SessionMeta` target guards against stale callbacks after SetSession swap
- Xcode.app path unaffected (no SetSession → no session → timer never arms)

Fixes the F1 gap surfaced by ACI-5045 manual test — wrapper builds now leave a pending record for F2's correlator to match.

Stacked on `aci-5044-slim-invocation-emit`.

## Test plan

- [x] `make lint` (0 issues)
- [x] `go test -tags unit -race ./internal/xcelerate/proxy/...` — new 5-case inactivity suite passes
- [ ] Manual re-run of ACI-5045 Scenario A after both PRs land — expect `matched=true` on wrapper build

### DIFF
--- a/cmd/xcode/xcelerate_start_proxy.go
+++ b/cmd/xcode/xcelerate_start_proxy.go
@@ -211,12 +211,18 @@ func newSlimInvocationEmitter(
 }
 
 func (e *slimInvocationEmitter) EmitSlim(ctx context.Context, meta proxy.SessionMeta, stats proxy.SessionStats) {
+	endTime := meta.EndTime
+	if endTime.IsZero() {
+		endTime = time.Now()
+	}
+	duration := endTime.Sub(meta.StartTime).Milliseconds()
+
 	// Fire-and-forget: proxy's shutdown/setSession critical path must not block on network I/O.
 	go func() {
 		inv := analytics.NewInvocation(analytics.InvocationRunStats{
 			InvocationDate: meta.StartTime,
 			InvocationID:   meta.InvocationID,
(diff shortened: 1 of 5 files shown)


Badlazzor:
## Manual test 2026-07-13

**F1 timer fires reliably:**

```
wrapper build: 12:23:55 → 12:24:00 (5s active)
F1 emit:       12:24:29 (30s after last RPC ✓)
pending row:   {"invocation_id":"5715289a-...","duration_ms":3736,"hit_rate":0}
```

Duration reports 3.7s (real RPC window) instead of 34s (wall clock at emit) — EndTime fix effective.

**Race exposed:** F2 processes manifest at t+5s (mints orphan `b5b71c0b-...`) before F1's pending lands at t+30s. Filed as **ACI-5210** — F2-side retry bucket is the preferred fix.

Badlazzor:
Merged into #423 (aci-5045-xcactivitylog-enrichment) — F1 inactivity-timer commits fast-forwarded onto that branch. Combined PR retitled/re-described.

PR #426, opened 2026-07-13
feat: ACI-5206 daemon writes pid file, wrapper reuses via shared helper

Badlazzor:
## Summary

- New `readAlivePid` + `acquireProxyPidLock` helpers in `cmd/xcode/pidlock.go`
- `xcelerate start-proxy` acquires the pid lock on startup; exits early with an "already running" log if a live proxy already holds it; releases on shutdown
- `startProxy` in the xcodebuild wrapper switches to `readAlivePid` — finds the daemon's pid, skips the subprocess spawn entirely
- Wrapper no longer writes the pid file itself (subprocess owns it)

Kills the duplicate-daemon symptom surfaced during the ACI-5045 manual test cycle — two `xcelerate start-proxy` processes bound to the same unix socket, second one rug-pulling the first's socket via `os.Remove`.

## Test plan

- [x] `make lint` (0 issues)
- [x] `go test -tags unit -race ./cmd/xcode/...` — new 6-case pidlock suite passes
- [x] Manual: LaunchAgent restart → wrapper build → `ps` shows only 1 `xcelerate start-proxy` bound to the socket; `~/.bitrise-xcelerate/proxy.pid` holds daemon's pid; subsequent wrapper invocations reuse


cmd/xcode/pidlock.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+// acquireProxyPidLock claims pidPath for the current process. Returns
+// errProxyAlreadyRunning if a different live process already holds it. Stale
+// or self-owned pid files are overwritten.
+//
+// The returned release closure removes the pid file — call it on shutdown.
+func acquireProxyPidLock(osProxy utils.OsProxy, pidPath string, isAlive processAliveFn) (func() error, error) {
+	pid, alive := readAlivePid(osProxy, pidPath, isAlive)
+	if alive && pid != os.Getpid() {
```
zsolt-marta-bitrise: Why is the `pid != os.Getpid()` needed?

cmd/xcode/pidlock.go
```diff
@@ -0,0 +1,79 @@
...
+	if err != nil {
+		return false
+	}
+
+	return proc.Signal(syscall.Signal(0)) == nil
+}
+
+// readAlivePid returns the pid recorded in pidPath and whether the process is
+// still running. Missing / malformed pid files return (0, false).
+func readAlivePid(osProxy utils.OsProxy, pidPath string, isAlive processAliveFn) (int, bool) {
```
zsolt-marta-bitrise: Can we reuse this in the doctor / status?

zsolt-marta-bitrise (approved):
Just some followup questions

cmd/xcode/pidlock.go
Badlazzor: Guards a same-process re-entry: if the current PID somehow re-enters `acquireProxyPidLock` (test reuse of a fake `osProxy` where the pid file survives, or a code path that re-invokes acquire without going through release), we want to reclaim, not error out. Dropping the check would treat our own pid as a foreign live owner and refuse the lock. Happy to inline a one-liner comment if useful.

cmd/xcode/pidlock.go
Badlazzor: Promoted to `internal/proxypid` in cc33862 (`Read` / `Acquire` / `ErrAlreadyRunning`). Doctor + daemon info already switched to socket probing in ab48558 (stronger signal than pid liveness) so those callers stay unchanged — pkg is available for any future consumer that needs the pid view.

zsolt-marta-bitrise:
[//]: # ([skip ci])



## Summary
Implements pid file locking to prevent duplicate daemon processes. The daemon acquires a lock on startup; the wrapper checks for an alive daemon via the pid file instead of spawning a new process. Eliminates the duplicate-daemon issue where concurrent wrapper invocations could start separate proxy processes.


## Walkthrough
| File | Summary |
|------|---------|
| internal/proxypid/proxypid.go | New pid file locking helpers: Read() checks if a process is alive, Acquire() claims pid file ownership with stale-file cleanup and idempotent release |
| cmd/xcode/xcelerate_start_proxy.go, cmd/xcode/xcodebuild.go | Daemon acquires pid lock on startup; wrapper uses Read() to detect alive daemon and reuses it instead of spawning subprocess; removed wrapper's pid file writing |
| internal/proxypid/proxypid_test.go | New test suite covering pid lock acquisition, stale file cleanup, and process liveness checks |

zsolt-marta-bitrise (approved):

PR #431, opened 2026-07-21
feat: ACI-5116 Bazel credential helper

Badlazzor:
## Summary
- Adds hidden root `bitrise-build-cache get` command implementing the [Bazel `--credential_helper` JSON protocol](https://bazel.build/reference/command-line-reference#flag--credential_helper). Reads `{"uri":...}` on stdin, resolves the auth token via the same precedence chain as the rest of the CLI (envs → keychain → multiplatform config), and writes `{"headers":{"authorization":["Bearer <token>"]}}` on stdout. Bazel spawns the helper as `<binary> get` per the EngFlow Credential Helper spec, so the entry point sits at the CLI root, hidden from `--help`. Root `PersistentPreRun` is bypassed on `get` for low-latency spawns.
- `activate bazel` (and `enable-for bazel`) now write `build --credential_helper=*.services.bitrise.io=<abs-cli-path>` into the local branch of `~/.bazelrc` instead of the two `--remote_header=authorization="Bearer <token>"` / `--bes_header=authorization=...` lines. Wildcard scopes the helper to Bitrise endpoints only, so Bazel does not call it for public URIs like `bcr.bazel.build`. CI activations (`CIProvider` set) keep the literal header — same split Gradle uses since ACI-5028: local dev keeps credentials out of plain text on disk, CI trades a per-RPC CLI spawn for stable header injection.
- Folds in a one-line comment accuracy fix in `cmd/common/root.go`: `hydrateStoredAuth` refreshes the PAT **in the credential store**, not "into the auth env vars".

Local snapshot from a smoke test (`activate bazel` with no `BITRISE_IO`):

```
build --credential_helper=*.services.bitrise.io=/tmp/bbc-aci-5116
build --remote_cache=grpcs://bitrise-accelerate.services.bitrise.io
build --remote_timeout=600s
build --remote_header=x-flare-buildtool=bazel
...
build --bes_backend=grpcs://flare-bes.services.bitrise.io:443
build --bes_results_url=https://app.bitrise.io/build-cache/invocations/bazel/
...
```

No literal `Bearer <token>` anywhere. The CI branch (`BITRISE_IO=true BITRISE_BUILD_SLUG=x`) still emits `--remote_header=authorization="Bearer <token>"` + `--bes_header=authorization=...`.

## Test plan
- [x] Unit tests: `internal/bazelcredhelper/` (JSON round-trip, malformed input, JWT/PAT parity with existing header behavior).
- [x] Unit tests: `internal/config/bazel/bazel_config_generate_test.go` — added `expectedLocalHelperConfig`, `expectedCIFallbackHeaders`, `expectedHelperCacheDisabled`; existing snapshots unchanged.
- [x] Manual smoke: `echo '{"uri":"x"}' | bitrise-build-cache get` returns valid JSON on stdout, logs stay on stderr.
- [x] Manual smoke: `activate bazel` on local (no `BITRISE_IO`) writes `--credential_helper=*.services.bitrise.io=<abs-path>`, no plain-text token.
- [x] Manual smoke: `activate bazel` with `BITRISE_IO=true BITRISE_BUILD_SLUG=x` keeps the literal `authorization="Bearer <token>"` lines.
- [x] E2E: new `feature-e2e-bazel-bitrisedc-credhelper` workflow (`bitrise.yml`) unsets `BITRISE_IO` to force the local-dev branch, runs `bazel build //src:bazel-dev` against `github.com/bitrise-io/bazel.git` using the generated `~/.bazelrc`, and asserts (a) `--credential_helper=*.services.bitrise.io=` is present, (b) no `Bearer ` / `authorization=` in `.bazelrc`, (c) the invocation reaches the API with `ciProvider=unknown`, `tool=bazel`, `cacheHitRate > 0`. Last run: 99.92% cache hit (3803/3806 cached actions) via the helper.


cmd/get/get.go
```diff
@@ -0,0 +1,42 @@
...
+// Credential Helper spec. This hidden root subcommand is the spawn target.
+
+//nolint:gochecknoglobals
+var getCmd = &cobra.Command{
+	Use:           "get",
+	Short:         "Bazel credential helper protocol entry point",
+	Hidden:        true,
+	SilenceUsage:  true,
+	SilenceErrors: true,
+	// Bazel spawns the helper N times in parallel with a tight per-invocation
```
zsolt-marta-bitrise, line 23: Nice thinking about the quick spawning

cmd/get/get.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/bitrise-io/bitrise-build-cache-cli/v3/internal/bazelcredhelper"
+	"github.com/bitrise-io/bitrise-build-cache-cli/v3/internal/utils"
+)
+
+// Bazel invokes `--credential_helper=<path>` as `<path> get`, per the EngFlow
+// Credential Helper spec. This hidden root subcommand is the spawn target.
+
+//nolint:gochecknoglobals
+var getCmd = &cobra.Command{
+	Use:           "get",
```
zsolt-marta-bitrise, line 18: This is so weird that we have a fixed argument here and have to add a non-sensible command here

internal/bazelcredhelper/helper.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+
+	cfg, _, err := configcommon.ResolveAuthConfig(envs)
+	if err != nil {
+		return fmt.Errorf("resolve auth config: %w", err)
+	}
+
+	resp := GetCredentialsResponse{
+		Headers: map[string][]string{
+			"authorization": {"Bearer " + cfg.AuthToken},
```
zsolt-marta-bitrise, line 54: can we read other locally static metadata like username/cpu/mem (paired with a backend change to parse them if anyone is missing, but afaik we have this already)? That way we would have that metadata without relying on customer wiring

zsolt-marta-bitrise (approved):

PR #434, opened 2026-07-22
feat: ACI-5095 refresh xcelerate proxy auth token via TTL-cached AuthProvider

Badlazzor:
## Problem

The Xcode xcelerate proxy resolves auth **once at startup** and hands the resulting `CacheAuthConfig` to a long-lived `kv.Client`. Every RPC inside that proxy session reuses the same in-memory token. If the PAT rotates mid-session (or the underlying keychain / multiplatform-credentials file is updated), the proxy must be restarted for the fresh token to take effect. Every other build tool (Gradle, Bazel, ccache, RN) already re-reads credentials on demand.

Tracked in ACI-5095.

## Design

**`configcommon.ExpiryAwareResolver`** — per-RPC re-read that routes OAuth-managed keychain reads through `oauth.EnsureFresh` (expiry-aware refresh + rotation), and falls through to the pure `ResolveAuthConfig` for env-var / file / multiplatform / JWT-fallback sources.

- New `kv.AuthSource` interface (single method: `Get() common.CacheAuthConfig`) is consumed by `kv.Client` on **every** RPC (via `getMethodCallMetadata`). Both the `authorization` header and the `x-org-id` header are read from the fresh value.
- `kv.NewClientParams.AuthSource` is optional. When nil, `NewClient` wraps the existing `AuthConfig` in a `staticAuthSource` — existing Gradle / Bazel / ccache / doctor / test call-sites keep working unchanged.
- `configcommon.ExpiryAwareResolver.Get()`:
  - Calls the pure `ResolveAuthConfig` first (env → keychain → file → multiplatform → JWT-fallback precedence).
  - If the resolved source is `AuthSourceKeychain` AND a `RefreshFunc` is wired in, invokes it (production wraps `oauth.Config.EnsureFresh`, which refreshes only within `refreshSkew` of `PATExpiry` and owns the refresh token via the store).
  - On refresh error: log-warn, serve the plain `ResolveAuthConfig` result (best-effort — downstream RPC will 401 if the token is truly dead).
- Xcode analytics client swaps its static `accessToken string` for a `tokenSupplier func() string` fed by the same resolver — invocation / cache-operation PUTs pick up rotated tokens too.

**Layering note**: the resolver uses a `RefreshFunc = func(ctx) (pat, wsid, error)` functional seam instead of a direct `oauth.Config` field, because `internal/oauth → internal/auth/store → internal/config/common` is a cycle. The closure wrapping `oauth.Config.EnsureFresh` lives at the `cmd/xcode/xcelerate_start_proxy.go` boundary where both packages are already imported.

## What is NOT touched

- Gradle, Bazel, RN, ccache paths — same behaviour as before (they wrap `AuthConfig` in the static source).
- On-disk config (`~/.bitrise-xcelerate/config.json`) — `AuthConfig` is still stripped at save.
- The CI-asserted log lines from `internal/ccache/*` (untouched).
- Non-OAuth sources fall through to plain `ResolveAuthConfig` — same behavior as before.

## Test plan

- [x] `make lint` clean.
- [x] `go test -tags unit -race ./...` green modulo the two pre-existing failures on `main` (`internal/xcelerate/enrichment/*` glob/orphan tests, and `pkg/ccache TestActivator_Activate/returns_error_when_auth_config_is_missing` when the dev keychain is populated).
- [x] `internal/config/common/expiry_aware_resolver_test.go` — non-OAuth source uses plain resolve, OAuth source uses refreshFn, refreshFn error falls back to plain cfg, nil refreshFn falls through, ResolveAuthConfig error returns cfg + log-warn.
- [x] `internal/build_cache/kv/authsource_internal_test.go` — `getMethodCallMetadata` returns fresh token/workspace on every call when the source rotates; stable when it doesn't.
- [x] Existing `Test_XcelerateProxy` assertion holds — proxy authorization header updates on every RPC.

## Review rounds

**Round 1 (Zsolt):**
- Thread `r3630875475` — `NewAuthProvider` signature drops the initial seed arg (commit `38ab2c1`)
- Thread `r3630881952` — resolver-error path no longer advances `fetchedAt` (commit `9206c97`)
- Thread `r3630901910` — TTL cache moved into `configcommon.CachingAuthResolver`, `internal/xcelerate/authprovider.go` deleted (commits `30f3ea1` + `ffe7e95`)
- Thread `r3630915325` — `feature-e2e-xcelerate-auth-refresh` workflow deleted (commit `92d651a`)

**Round 2 (Zsolt, comment `r3637034267`):**
- TTL cache dropped entirely — `ExpiryAwareResolver` calls per-RPC through `oauth.EnsureFresh` for OAuth-managed sources (commit `b828d97`). Real expiry-aware refresh replaces the 60s blind timer.


internal/xcelerate/authprovider.go
```diff
@@ -0,0 +1,92 @@
...
+	mu        sync.RWMutex
+	cached    common.CacheAuthConfig
+	seeded    bool
+	fetchedAt time.Time
+	nowFn     func() time.Time
+}
+
+// NewAuthProvider constructs a provider seeded with initial so the first Get
+// returns without calling resolver. Subsequent Gets refresh once ttl elapses.
+func NewAuthProvider(initial common.CacheAuthConfig, resolver AuthResolver, ttl time.Duration, logger log.Logger) *AuthProvider {
```
zsolt-marta-bitrise: I don't see why we pass both the authconfig and the authresolver. Can we just pass the resolver and call get on it? Would simplify this

internal/xcelerate/authprovider.go
```diff
@@ -0,0 +1,92 @@
...
+	if p.resolver == nil {
+		return p.cached
+	}
+
+	fresh, err := p.resolver()
+	if err != nil {
+		if p.logger != nil {
+			p.logger.Warnf("xcelerate auth token refresh failed, using cached value: %s", err)
+		}
+		p.fetchedAt = p.nowFn()
```
zsolt-marta-bitrise: In case of error I don't think we should save the fetchedAt, our token might be stale

internal/xcelerate/authprovider.go
```diff
@@ -0,0 +1,92 @@
...
+	fetchedAt time.Time
+	nowFn     func() time.Time
+}
+
+// NewAuthProvider constructs a provider seeded with initial so the first Get
+// returns without calling resolver. Subsequent Gets refresh once ttl elapses.
+func NewAuthProvider(initial common.CacheAuthConfig, resolver AuthResolver, ttl time.Duration, logger log.Logger) *AuthProvider {
+	p := &AuthProvider{
+		resolver: resolver,
+		ttl:      ttl,
```
zsolt-marta-bitrise: Why is this handled here and not in the auth resolver? The access token can be read as many times as we want, why do we guard `configcommon.ResolveAuthConfig`? If we are guardign the backend, isn't it the job of configcommon to do so because ultimately it decides if we need to refresh the token (it should have access to the refresh token anyways)

bitrise.yml
```diff
@@ -1370,6 +1370,97 @@ workflows:
             assert_xcode_no_orphan_invocations_for_build "$BITRISE_BUILD_SLUG" "$INVOCATION_ID"
     - deploy-to-bitrise-io: {}
 
+  feature-e2e-xcelerate-auth-refresh:
+    description: |
+      ACI-5095: proves the AuthProvider 60s TTL refresh path actually fires in
```
zsolt-marta-bitrise: I think this is a bit of an overkill for an e2e test, especially because we are depending on internal logs to verify instead of real token validity. Unit tests shuld be fine (and we're getting these xcode tests a bit too bloated). So either test token expiration (not sure if we can mint very short tokens), or rely on unit test

internal/xcelerate/authprovider.go
Badlazzor: right, sounds better 👍

internal/xcelerate/authprovider.go
Badlazzor: Done in 38ab2c1 — dropped the `initial` param; first `Get()` calls the resolver under write-lock. Same behavior end-to-end, cleaner surface.

internal/xcelerate/authprovider.go
Badlazzor: Good catch. Fixed in 9206c97 — `fetchedAt` no longer updates on resolver error, so the next `Get()` retries within the TTL window instead of waiting the full 60s. New test `TestAuthProvider_ResolverErrorRetriesNextCall` covers it.

internal/xcelerate/authprovider.go
Badlazzor: Agreed and done in 30f3ea1 + ffe7e95. New `configcommon.CachingAuthResolver` owns the TTL cache. Xcelerate proxy uses it; `internal/xcelerate/authprovider.go` deleted. Kept the pure `ResolveAuthConfig` untouched — the ~25 one-shot callers (command entry points) don't want stateful caching. When we add JWT refresh later it'll slot into the same `configcommon` layer.

bitrise.yml
Badlazzor: Agreed. Deleted the workflow in 92d651a. Unit tests cover both refresh-after-TTL and error-retry-next-call, so nothing lost.

zsolt-marta-bitrise:
[//]: # ([skip ci])



## Summary
Implement per-RPC auth refresh for Xcode xcelerate proxy via ExpiryAwareResolver. Routes OAuth-managed keychain reads through expiry-aware refresh while falling back to plain ResolveAuthConfig for other sources. Enables token rotation mid-session without proxy restart.


## Walkthrough
| File | Summary |
|------|---------|
| internal/config/common/expiry_aware_resolver.go, expiry_aware_resolver_test.go | New ExpiryAwareResolver that calls RefreshFunc for OAuth keychain sources, falls back to plain ResolveAuthConfig for env/file/multiplatform/JWT sources, and uses context.WithoutCancel to decouple from signal context. |
| internal/build_cache/kv/client.go, methods.go, authsource_internal_test.go | Introduce AuthSource interface for per-RPC credential retrieval. Client calls authSource.Get() on every RPC via getMethodCallMetadata. Backward compatible: nil AuthSource wraps AuthConfig in staticAuthSource. |
| cmd/common/client.go | Add optional AuthSource parameter to CreateKVClientParams, passed through to kv.NewClient for per-RPC refresh. |
| cmd/xcode/xcelerate_start_proxy.go | Wire ExpiryAwareResolver as AuthSource to kv.Client and analytics bundle for per-RPC token refresh. Use context.WithoutCancel to prevent signal context cancellation. |
| cmd/xcode/restoreXcodeDerivedDataFiles.go, saveXcodeDerivedDataFiles.go, xcodebuild.go | Update analytics client to use tokenSupplier func instead of static accessToken string. |
| internal/xcelerate/analytics/client.go, cache_operations.go, invocations.go | Replace static accessToken with tokenSupplier func() string, called on every request for fresh credentials. |
| internal/config/common/benchmark.go, benchmark_test.go | Add Getenv field to BenchmarkPhaseClient for test injection and env var override support. |

internal/xcelerate/authprovider.go
zsolt-marta-bitrise: Thanks — but stepping back, I'm not sure the TTL cache is the right primitive here, and I don't think we need this class at all.

As wired, `CachingAuthResolver` wraps `ResolveAuthConfig`, which is a pure local re-read (env → keychain → file → multiplatform). It refreshes nothing — `internal/oauth.EnsureFresh` is never called on this path (only `auth login` calls it). So two things:

1. It doesn't actually solve the mid-session rotation problem for OAuth tokens. The resolver just re-reads whatever PAT is already in the keychain; a rotation only propagates if something *else* rewrote the store. The TTL re-reads, it doesn't refresh.
2. Since the wrapped resolve is cheap local IO, the only thing the 60s TTL buys is avoiding those reads — at the cost of up to 60s staleness. For rotation pickup we could just call `ResolveAuthConfig` per-RPC (same as the other tool paths) with zero staleness and no new class.

The expiry-aware logic we actually want already exists one layer down: `EnsureFresh` refreshes only within `refreshSkew` of `PATExpiry`, and it owns the refresh token. That's the "configcommon decides when to refresh" I was after. A blind 60s timer is a worse proxy for it — it either refreshes needlessly while the token is valid for hours, or serves an expired token for up to 60s.

Proposal: drop the TTL class. Route the resolver through the expiry-aware path — call `EnsureFresh` when the credential is OAuth-managed keychain, fall through to the plain re-read otherwise — and have the proxy call the resolver per-RPC. Instant rotation pickup, network refresh only near expiry, nothing arbitrary to reason about.

Badlazzor: Agreed and done in `b828d97`. `CachingAuthResolver` + tests deleted. New `configcommon.ExpiryAwareResolver` routes OAuth-managed keychain reads through a refresh closure (production wraps `oauth.Config.EnsureFresh`) and falls through to the pure `ResolveAuthConfig` for env-var / file / multiplatform / JWT-fallback sources. Proxy calls `Get()` per RPC — no TTL, no staleness, no arbitrary timer.

Used a functional seam (`RefreshFunc = func(ctx) (pat, wsid, error)`) instead of a direct `oauth` import from `configcommon` because `internal/oauth → internal/auth/store → internal/config/common` is a cycle. The closure wrapping lives at the `cmd/xcode` boundary where both packages are already imported.

internal/config/common/expiry_aware_resolver.go
```diff
@@ -0,0 +1,60 @@
...
+)
+
+// RefreshFunc plugs a live credential source (e.g. oauth.Config.EnsureFresh) into ExpiryAwareResolver without pulling internal/oauth into this package (would cycle via internal/auth/store).
+type RefreshFunc func(ctx context.Context) (pat string, workspaceID string, err error)
+
+type resolveFunc func(envs map[string]string) (CacheAuthConfig, AuthSource, error)
+
+// ExpiryAwareResolver routes OAuth-managed keychain reads through refreshFn (expiry-aware refresh + rotation) and falls back to plain ResolveAuthConfig for env / file / multiplatform / JWT sources.
+type ExpiryAwareResolver struct {
+	ctx       context.Context //nolint:containedctx // resolver is called per RPC without a fresh ctx
```
zsolt-marta-bitrise, line 16: [bitrise-plugin-ai-reviewer]: internal/config/common/expiry_aware_resolver.go:16:0:b828d97a

[//]: # ([skip ci])


**🐛 Bug**

The stored context is the proxy's signal context (created via signal.NotifyContext). During graceful shutdown, this context will be cancelled, causing all subsequent refresh calls to fail immediately. The resolver should use context.Background() or a fresh context for each refresh call instead of the stored signal context.

<details>

```
In internal/config/common/expiry_aware_resolver.go at line 16, Use context.Background() for refresh calls instead of the stored signal context. The signal context should only control proxy shutdown, not individual RPC credential refreshes.
```

</details>

cmd/xcode/xcelerate_start_proxy.go
```diff
@@ -162,10 +163,23 @@ func StartXcodeCacheProxy(
...
+	oauthCfg.Logger = initialLogger
+	refreshFn := func(ctx context.Context) (string, string, error) {
+		creds, err := oauthCfg.EnsureFresh(ctx)
+		if err != nil {
+			return "", "", fmt.Errorf("ensure fresh oauth credentials: %w", err)
+		}
+
+		return creds.PAT, creds.WorkspaceID, nil
+	}
+	authProvider := configcommon.NewExpiryAwareResolver(ctx, envProvider, refreshFn, initialLogger)
```
zsolt-marta-bitrise: [bitrise-plugin-ai-reviewer]: cmd/xcode/xcelerate_start_proxy.go:176:0:b828d97a

[//]: # ([skip ci])


**🐛 Bug**

The ExpiryAwareResolver is initialized with `ctx` (which is `signalCtx` from line 117). This context will be cancelled when a SIGTERM or SIGINT signal is received. After cancellation, all subsequent RPC calls will attempt to use the cancelled context in the `refreshFn` closure (line 169), causing all auth refresh attempts to fail with "context cancelled" errors. The proxy should use a non-cancellable context for the resolver to ensure auth refreshes work throughout the proxy's lifetime, even after receiving shutdown signals.

🔄 Suggestion:
```suggestion
	authProvider := configcommon.NewExpiryAwareResolver(context.WithoutCancel(ctx), envProvider, refreshFn, initialLogger)
```

zsolt-marta-bitrise (commented):

zsolt-marta-bitrise (approved):

PR #436, opened 2026-07-22
feat: ACI-5227 explicit EndSession RPC + F2 manifest span as authoritative duration

Badlazzor:
Ticket: https://bitrise.atlassian.net/browse/ACI-5227
Follow-up to Zsolt's review on #423: https://github.com/bitrise-io/bitrise-build-cache-cli/pull/423#discussion_r3614707559

## Summary

Two coupled changes so the enrichment watcher's re-PUT is the authoritative source of Xcode invocation duration **for wrapper-less builds**, without disturbing the wrapper's own final duration for wrapper builds.

### Axis 1 — `Session.EndSession` RPC (deterministic build-done signal)

- New `EndSession(EndSessionRequest{invocation_id, end_time_unix_ms})` RPC on `Session`.
- Proxy flushes the current session immediately (with the wrapper-supplied end_time_unix_ms) instead of waiting on the inactivity timer.
- Wrapper (`cmd/xcode/xcodebuild.go`) calls `EndSession` right after `XcodeRunner.Run` returns and before analytics is saved. Failure is debug-logged only — never fails analytics.
- Mismatched or absent invocation ID = silent no-op + debug log (no gRPC error).
- `defaultInactivityTimeout` bumped `10 * time.Second` → `5 * time.Minute` so long xcodebuild silent phases (Swift concurrency stalls, dependency resolution) don't trip a premature slim emit.

### Axis 2 — manifest span as authoritative duration (wrapper-less path only)

- `slimInvocationEmitter.EmitSlim` no longer sends `Duration` on the slim PUT payload. The enrichment watcher re-PUTs with `entry.Stop - entry.Start` from `LogStoreManifest` which is the true build span.
- `PendingRecord.Duration` stays — the correlator still uses it as a time-overlap window when matching manifests to pending records.
- Wrapper's own analytics PUT (via `saveInvocationAndRelation`) keeps its wall-clock Duration. The handled-invocation marker means the watcher's re-PUT skips wrapper rows, so the wrapper's duration is final for wrapper builds.

### Which duration lands where

Manifest-span authoritative duration only takes effect for wrapper-less builds. For wrapper builds the wrapper's rich PUT is the only surviving row (the marker makes both the slim emit and the watcher skip), and its wall-clock `runStats.DurationMS` is the final duration.

| Build type   | ID                   | Duration                        | Hit rate                              |
|--------------|----------------------|---------------------------------|---------------------------------------|
| Wrapper      | `SetSession` ID      | wrapper wall-clock — final      | log-scraped target rate — final       |
| Wrapper-less | watcher-minted UUID  | manifest span — final           | (proxy has none; watcher's)           |

## Test plan

- [x] `TestSetSession_EndSession_emitsSingleInvocation` — SetSession → EndSession → exactly one emit
- [x] `TestSetSession_EndSession_prevents_inactivity_double_emit` — EndSession cancels the pending inactivity timer
- [x] `TestSetSession_EndSession_usesSuppliedEndTime` — `meta.EndTime` reflects the wrapper-supplied `end_time_unix_ms`
- [x] `TestEndSession_UnknownInvocation_NoOp` — mismatched invocation id does not flush the active session
- [x] `TestEndSession_NoActiveSession_NoOp` — no panic when EndSession is called without a session
- [x] `Test_xcodebuildCmdFn_callsEndSessionAfterRun` — wrapper invokes EndSession exactly once with matching id and current wall-clock
- [x] `Test_slimInvocationEmitter_EmitSlim_omitsDurationOnPUT` — via new `invocationSaver` seam on analyticsBundle
- [x] `Test_slimInvocationEmitter_EmitSlim_persistsPendingRecord` — preserved: pending Duration still 2500ms
- [x] `Test_slimInvocationEmitter_EmitSlim_skipsWhenMarkerPresent` — preserved: append-before-marker invariant
- [x] `TestEnricher_DurationIsFromManifest` — watcher PUT `DurationMs = 42_000` from manifest span, not pending's provisional 2500

## Notes

- **Test-seam naming deviation:** plan called for a new `invocationPutter` interface on `analyticsBundle`. `cmd/xcode/xcodebuild.go` already exposes an identical `invocationSaver` interface — lint (iface/identical) flagged the duplicate, so `analyticsBundle.putter` reuses `invocationSaver` instead of adding a duplicate. Same semantics; test still uses a capturing struct that satisfies the interface.
- `internal/xcelerate/proxy/inactivity_test.go::TestInactivityDuration_ZeroFallsBackToDefault` updated to assert `5 * time.Minute`.
- Unrelated protoc-generated files (`kv_storage`, `remote_execution`, `llvm/cas`, `llvm/kv`) were reverted to keep the diff scoped — only `proto/llvm/session/*` is touched.
- Verified locally: `make lint` clean, targeted `go test -tags unit -race ./cmd/xcode/... ./internal/xcelerate/proxy/... ./internal/xcelerate/enrichment/ -run TestEnricher_` green. Pre-existing flakes in `enrichment/watcher_internal_test.go`, `enrichment/retry_test.go`, `pkg/ccache`, `internal/config/{gradle,bazel}`, `cmd/{gradle,bazel}` all present on unchanged main.
- Follow-up commit renamed the F1/F2 shorthand across code + docs + bitrise.yml to `wrapper` / `watcher` (Zsolt's review comment on `xcelerate_start_proxy.go:370`).


cmd/xcode/xcelerate_start_proxy.go
```diff
@@ -358,15 +361,20 @@ func (e *slimInvocationEmitter) EmitSlim(ctx context.Context, meta proxy.Session
 		return
 	}
 
+	putter := b.resolvePutter()
+	if putter == nil {
+		return
+	}
+
 	go func() {
+		// Duration omitted: F2's manifest span or the wrapper's own PUT is authoritative.
```
zsolt-marta-bitrise: I just realized that F1/F2 won't mean anything outside the context of these PRs. Can you please remove references to them, it's not a real feature just an abbreviation to understand what's happening. We can call them wrapper / watcher

internal/xcelerate/proxy/proxy.go
```diff
@@ -223,6 +223,35 @@ func (p *Proxy) onInactivity(target *SessionMeta) {
...
+		p.logger.TDebugf("EndSession no-op: got %s, active %s", req.GetInvocationId(), p.currentSession.InvocationID)
+
+		return &emptypb.Empty{}, nil
+	}
+
+	if endMs := req.GetEndTimeUnixMs(); endMs > 0 {
+		p.lastActivity = time.UnixMilli(endMs)
+	}
+
+	p.emitCurrentSessionLocked(ctx)
```
zsolt-marta-bitrise, line 250: Do we / should we clear the current session?

cmd/xcode/xcodebuild.go
```diff
@@ -393,6 +393,16 @@ func (c *XcodebuildRunner) Run(ctx context.Context) xcodeargs.RunStats {
 	c.appendLocalInvocationLog(*inv, runStats)
 	c.saveInvocationAndRelation(*inv, runStats.CacheStats.Hits, runStats.CacheStats.TotalTasks)
 
+	// Signal build-done AFTER the wrapper's own PUT + marker write so the proxy's slim emit sees the marker and skips.
```
zsolt-marta-bitrise, line 396: I had to think a bit how the marker works in this case now

 **On the marker + watcher "authoritative duration" framing**                                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                                                            
  Traced the full flow — I think the marker is fine to keep (it's the existing mechanism), but the PR's duration framing doesn't match what actually happens for wrapper builds. Two facts that matter:                                                                                                                     
                                                                                                                                                                                                                                                                                                                            
  1. **`SetSession` is wrapper-only.** `p.currentSession` is created solely in `SetSession` (`proxy.go:225`), and its only caller is the wrapper (`xcodebuild.go:349`). Wrapper-less builds (IDE / direct `xcodebuild`) never open a proxy session, so they get no slim emit and no pending record — their sole analytics   
  path is the watcher minting a UUID off the `LogStoreManifest` (enricher orphan path). The slim-emit + pending machinery is wrapper-only.                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                            
  2. **The marker makes both the slim emit and the watcher skip for wrapper builds.** So the wrapper's own rich PUT is the *only* surviving row for a wrapper build — the watcher's manifest-span duration never lands there.                                                                                               
                                                                                                                                                                                                                                                                                                                            
  Consequence: **the manifest-span authoritative duration only takes effect for wrapper-less builds.** For wrapper builds the authoritative duration is the wrapper's `runStats.DurationMS` (wall-clock), which the rich PUT already carries (`xcodebuild.go:394`) — this is deliberate and correct, since the watcher can't
  reconstruct the log-scraped target hit rate that the wrapper owns, so we don't want the watcher overwriting wrapper rows anyway.                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                            
  So no code change needed on the duration itself — just the wording:                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                            
  - The wrapper's duration isn't "provisional" — under the marker, nothing overwrites it, so it's final for wrapper builds.                                                                                                                                                                                                 
  - "re-PUT overwrites when correlated" / "authoritative end-to-end" is only true for the wrapper-less path. Suggest restating as "manifest span is authoritative for wrapper-less builds; wrapper builds keep their own wall-clock duration."                                                                              
                                                                                                                                                                                                                                                                                                                            
  Net split, for the description:                                                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                                            
  | Build type | ID | Duration | Hit rate |                                                                                                                                                                                                                                                                                 
  |---|---|---|---|                                                                                                                                                                                                                                                                                                         
  | Wrapper | `SetSession` ID | wrapper wall-clock — final | log-scraped target rate — final |                                                                                                                                                                                                                              
  | Wrapper-less | watcher-minted UUID | manifest span — final | (proxy has none; watcher's) |

cmd/xcode/xcelerate_start_proxy.go
Badlazzor: Yeah, was planning to

Badlazzor: Renamed F1/F2 → `wrapper` / `watcher` across code, tests, `bitrise.yml` (including the workflow slug `e2e-xcode-f1-f2-race` → `e2e-xcode-wrapper-watcher-race`), and the July 2026 xcode-app findings doc. Zero remaining `F1`/`F2` references. Fix in f57b544.

internal/xcelerate/proxy/proxy.go
Badlazzor: It is cleared — indirectly. `emitCurrentSessionLocked` (proxy.go:148) nils `p.currentSession` at line 164 on its way out, and `EndSession` funnels through that same helper (line 250). Same clearing path as the inactivity-timer and `SetSession` swap flows, so the invariant is: any successful emit leaves `currentSession == nil`. No separate clear needed in `EndSession`.

cmd/xcode/xcodebuild.go
Badlazzor: PR body updated with the two-row split table and the reworded framing (manifest span is authoritative for wrapper-less builds; wrapper builds keep their wall-clock duration). Also reworded the in-code comment at `cmd/xcelerate_start_proxy.go:370` — the slim PUT drops Duration because "manifest span (from the enrichment watcher, wrapper-less builds only) or the wrapper's own PUT is authoritative". Both landed alongside the F1/F2 rename in f57b544.

zsolt-marta-bitrise:
[//]: # ([skip ci])



## Summary
Implements explicit EndSession RPC and manifest span as authoritative duration for wrapper-less builds. Wrapper calls EndSession after analytics save to signal build completion. Enrichment watcher re-PUTs with manifest-derived duration for wrapper-less builds only; wrapper builds retain their wall-clock duration. Inactivity timeout bumped to 5 minutes. Terminology refactored from F1/F2 to wrapper/watcher.


## Walkthrough
| File | Summary |
|------|---------|
| proto/llvm/session/session.proto, session.pb.go, session_grpc.pb.go | Added EndSession RPC method and EndSessionRequest message with invocation_id and end_time_unix_ms fields |
| internal/xcelerate/proxy/proxy.go | Added EndSession RPC handler that flushes current session if invocation ID matches; silent no-op for mismatched/absent sessions. Bumped default inactivity timeout from 10s to 5 minutes |
| internal/xcelerate/proxy/end_session_test.go | Added 5 tests covering EndSession single emit, inactivity timer cancellation, supplied end time, unknown invocation, and no-active-session scenarios |
| internal/xcelerate/proxy/inactivity_test.go | Updated test assertion for new 5-minute default inactivity timeout |
| cmd/xcode/xcodebuild.go | Wrapper calls EndSession RPC after saveInvocationAndRelation with current wall-clock time |
| cmd/xcode/xcodebuild_test.go, xcodebuild_internal_test.go | Added test for EndSession invocation with correct ID and timing; updated mock to satisfy SessionClient interface |
| cmd/xcode/xcelerate_start_proxy.go | Added putter field to analyticsBundle for test injection; EmitSlim now omits Duration from slim PUT (set to 0) |
| cmd/xcode/xcelerate_start_proxy_internal_test.go | Added test verifying Duration is omitted from slim PUT while pending records retain it |
| internal/xcelerate/enrichment/enricher.go, enricher_test.go | Duration now sourced from manifest entry (Stop - Start) instead of pending record. Added test verifying manifest span is authoritative |
| internal/xcelerate/enrichment/handled_marker.go, pending.go, invocation_emit.go | Updated comments to clarify enrichment watcher re-PUT skips wrapper rows via handled-invocation marker |
| internal/config/common/benchmark.go, benchmark_test.go | Added Getenv field to BenchmarkPhaseClient for environment variable override; checks pre-set BITRISE_BUILD_CACHE_BENCHMARK_PHASE_XCODE before API query |
| bitrise.yml | Renamed e2e-xcode-f1-f2-race workflow to e2e-xcode-wrapper-watcher-race; added BITRISE_BUILD_CACHE_BENCHMARK_PHASE_XCODE: established to prevent baseline detection |
| internal/config/xcelerate/paths.go, internal/paths/paths.go, docs/xcode-app-ide-remote-cas-findings-2026-07-21.md | Terminology refactored from F1/F2 to wrapper/watcher across comments and documentation |

Badlazzor:
CI red on `e2e-xcode-wrapper-watcher-race` was caused by the workflow rename in f57b544: the BE benchmark-phase endpoint keys on `workflow_name`, so it treated the renamed slug as brand-new -> returned `baseline` -> cache disabled -> `cacheHitRate=0` -> the workflow's hit-rate assertion failed.

Resolved by backporting the benchmark env-var override from PR #434 (commit c7d2a0c on `aci-5095-xcelerate-auth-refresh`): `BenchmarkPhaseClient.GetBenchmarkPhase` now short-circuits to `BITRISE_BUILD_CACHE_BENCHMARK_PHASE_<TOOL>` when it's already set, skipping the BE query. Set the env to `established` on `e2e-xcode-wrapper-watcher-race` so the cache stays on and the assertion is meaningful.

Pushed as c7c9f1a.

cmd/xcode/xcodebuild.go
zsolt-marta-bitrise: Nice!

zsolt-marta-bitrise (approved):

PR #442, opened 2026-07-27
fix: SSW-3090 make gradle-mirrors init script Isolated-Projects-compatible

lpusok:
## Summary

[SSW-3090](https://bitrise.atlassian.net/browse/SSW-3090) — Adastra s.r.o. enabled Gradle's Isolated Projects (`org.gradle.unsafe.isolated-projects=true`) and every build fails at configuration time:

```
Plugin class 'Bitrise_gradle_mirrors_init_gradle$InternalRepositoryPlugin':
Project ':' cannot access 'Project.tasks' functionality on subprojects via 'allprojects'
Plugin class 'Bitrise_gradle_mirrors_init_gradle$InternalRepositoryPlugin':
Project ':build-logic' cannot access 'Project.tasks' functionality on subprojects via 'allprojects'
```

The Robolectric block in `gradle.afterProject` reached into every project's `tasks` via `getRootProject().allprojects`. It is gated on `UseAsRobolectricRepo`, set only on the `mavencentral` mirror — which is on by default, so any Isolated Projects build hits it. This was the only `allprojects`/`subprojects` usage in the repo; the build-cache init script uses `rootProject { }` only and is unaffected. Note the script also ships via preboot, so the blast radius is not limited to workflows running the Activate Gradle Mirrors step.

## Changes

Both in `asset/gradle-mirrors.init.gradle.kts.gotemplate`:

- **Register the Robolectric redirect per project, in `beforeProject`.** Drops the `getRootProject().allprojects { }` wrapper, which is the Isolated Projects violation. `configureEach` stays lazy, so coverage is unchanged.

  `beforeProject` rather than `afterProject` is deliberate, and it is the non-obvious part of this PR. `configureEach` actions run in registration order, so whoever registers last wins. The old `allprojects`-from-root registration happened *before* subproject build scripts were read, so a project setting `robolectric.dependency.repo.url` itself won. Moving to each project's own `afterProject` would invert that and make us the last writer — silently overriding the escape hatch we published on [SSW-3019](https://bitrise.atlassian.net/browse/SSW-3019):

  ```kotlin
  tasks.withType<Test>().configureEach {
      systemProperty("robolectric.dependency.repo.url", "https://repo1.maven.org/maven2")
  }
  ```

  `beforeProject` registers ahead of the build script, so project config wins again — now consistently, including the root project, which the old code did not manage.

- **Suffix `roboProbes` per mirror** (`roboProbes{{ .ID }}`), matching the existing `canBeMirrored{{ .ID }}` / `useMirror{{ .ID }}` convention. Only `mavencentral` sets `UseAsRobolectricRepo` today, so one copy is emitted; a second flagged entry would otherwise emit two `val roboProbes` in one scope, fail to compile, and break every Gradle build on the machine.

- **Take the `Test` task from the `doFirst` receiver** instead of capturing it. `RobolectricDiagnostics.probe` put a `Task` reference inside its own task action; Isolated Projects implies the configuration cache, so that would have been the next failure once the `allprojects` one cleared.

Plus a log reword (the root-only line described work that now happens per project) and ASCII-only in the template.

`activate_test.go` — update the pinned assertions and add `allprojects` / `subprojects` / `getRootProject()` to `expectNotContain` as a regression guard.

## Verification

Multi-module repro with a `build-logic` included build, `org.gradle.unsafe.isolated-projects=true`, Gradle 8.14.5 / JDK 21. Rendered with the full `KnownMirrors` set.

| Check | Result |
| --- | --- |
| Pre-fix script | reproduces the customer error verbatim |
| Post-fix, probes on, no override | mirror applies, no IP error, config cache **stored** |
| Re-run | `Reusing configuration cache.` |
| SSW-3019 escape hatch present | resolves to `repo1.maven.org` — project config wins |
| `BITRISE_MAVENCENTRAL_PROXY_ENABLED=false` | script no-ops, build green |
| `BITRISE_ROBOLECTRIC_JAR_AUDIT=false` | green |
| Static | 0 non-ASCII, 0 `allprojects`/`getRootProject()`, exactly 1 `roboProbes` decl |

`make lint` → 0 issues. Full `make test-unit` green except `internal/xcelerate/enrichment` and `pkg/ccache`, which fail identically on clean `main`.

## Notes for the reviewer

See also the [follow-up comment](https://github.com/bitrise-io/bitrise-build-cache-cli/pull/442#issuecomment-5089912615) on retiring `RobolectricDiagnostics` and on the root-project `testLogging` precedence shift.

Worth a separate ticket: migrating `beforeProject`/`afterProject` to the isolated `gradle.lifecycle.*` API would be fully IP-clean, but those actions cannot capture `log()`, the `configureMirror`/`useMirror` `Action`s, or the `loggedReplacements` dedupe set — a rewrite of the whole script plus a Gradle 8.8 floor. Per [gradle/gradle#30075](https://github.com/gradle/gradle/issues/30075) the current hooks are "not inherently incompatible with IP", so there is no correctness pressure to do it now.

[SSW-3090]: https://bitrise.atlassian.net/browse/SSW-3090?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SSW-3019]: https://bitrise.atlassian.net/browse/SSW-3019?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/config/gradle/mirrors/activate_test.go
+++ b/internal/config/gradle/mirrors/activate_test.go
@@ -81,7 +81,8 @@ func TestActivate(t *testing.T) {
 				`log("prepending pluginManagement mirror https://repository-manager.services.bitrise.io:8090/maven/gradle-plugins")`,
 				`val loggedReplacements = java.util.concurrent.ConcurrentHashMap.newKeySet<String>()`,
 				`if (loggedReplacements.add("${getName()}|${getUrl()}|$mirrorUrl"))`,
-				// Core: robolectric android-all redirect set unconditionally as part of the mirror install
+				// Core: robolectric android-all redirect set unconditionally as part of the mirror install,
+				// configured per project so it stays Isolated-Projects-compatible
 				`log("setting robolectric.dependency.repo.url=\"https://repository-manager.services.bitrise.io:8090/maven/central\" on all Test tasks")`,
 				`val roboProbes = System.getenv("BITRISE_ROBOLECTRIC_JAR_AUDIT") != "false"`,
 				`tasks.withType(Test::class.java).configureEach {`,
@@ -94,7 +95,7 @@ func TestActivate(t *testing.T) {
 				`events("failed")`,
 				// Robolectric-specific classpath probe
 				`object RobolectricDiagnostics {`,
-				`test.doFirst { emitClasspathProbe(test, cliVersion) }`,
+				`test.doFirst { emitClasspathProbe(this as Test, cliVersion) }`,
 				// Classpath fingerprint (noncompatWidthPixels field check + bytecode-agent inventory, robolectric#9630)
 				`val hasField = String(bytes, Charsets.ISO_8859_1).contains("noncompatWidthPixels")`,
 				`"${jar.name}(${jar.length()},noncompatWidthPixels:$field)"`,
@@ -119,6 +120,10 @@ func TestActivate(t *testing.T) {
 				"getJavaLauncher",
 				"BITRISE_ROBOLECTRIC_VERBOSE_LOGGING",
 				"finalizedBy",
+				// Cross-project access — breaks Gradle Isolated Projects
+				"allprojects",
+				"subprojects",
+				"getRootProject()",
 			},
 		},
 		{
--- a/internal/config/gradle/mirrors/activate_test.go
+++ b/internal/config/gradle/mirrors/activate_test.go
@@ -83,10 +83,13 @@ func TestActivate(t *testing.T) {
 				`if (loggedReplacements.add("${getName()}|${getUrl()}|$mirrorUrl"))`,
 				// Core: robolectric android-all redirect set unconditionally as part of the mirror install,
 				// configured per project so it stays Isolated-Projects-compatible
-				`log("setting robolectric.dependency.repo.url=\"https://repository-manager.services.bitrise.io:8090/maven/central\" on all Test tasks")`,
-				`val roboProbes = System.getenv("BITRISE_ROBOLECTRIC_JAR_AUDIT") != "false"`,
+				`log("registering robolectric.dependency.repo.url=\"https://repository-manager.services.bitrise.io:8090/maven/central\" on Test tasks, per project")`,
+				// Suffixed per mirror: two UseAsRobolectricRepo entries would otherwise emit conflicting
+				// declarations in the same scope and the init script would not compile
+				`val roboProbesCentral = System.getenv("BITRISE_ROBOLECTRIC_JAR_AUDIT") != "false"`,
 				`tasks.withType(Test::class.java).configureEach {`,
 				`systemProperty("robolectric.dependency.repo.url", "https://repository-manager.services.bitrise.io:8090/maven/central")`,
+				`if (roboProbesCentral) {`,
 				`TestFailureLogging.enable(this)`,
 				`RobolectricDiagnostics.probe(this, "v0.0.0-test")`,
 				// Generic FULL-on-FAILED logging extracted into its own object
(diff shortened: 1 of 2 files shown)


zsolt-marta-bitrise:
[//]: # ([skip ci])



## Summary
Fix Gradle Isolated Projects compatibility by eliminating cross-project access patterns in the gradle-mirrors init script. Move Robolectric configuration to `beforeProject` and fix configuration cache serialization by capturing task context instead of task references.


## Walkthrough
| File | Summary |
|------|---------|
| gradle-mirrors.init.gradle.kts.gotemplate | Removed getRootProject().allprojects wrapper; moved Robolectric config to beforeProject. Changed task action to capture context (this as Test) instead of Task reference for configuration cache compatibility. Added variable suffixes to prevent mirror conflicts. |
| activate_test.go | Updated assertions for new log messages and variable names. Added regression guards for cross-project access patterns (allprojects, subprojects, getRootProject()). |

zsolt-marta-bitrise (approved):

lpusok:
Two notes for review, both prompted by re-reading [zsolt-marta-bitrise's write-up on robolectric#10586](https://github.com/robolectric/robolectric/issues/10586#issuecomment-4864155400) after a code-review pass. Neither changes the code in this PR — they're context for deciding what to keep.

### 1. The `doFirst` change is contingent on `RobolectricDiagnostics` surviving

The second half of this PR (taking the task from the `doFirst` receiver instead of capturing it) exists *only* because `RobolectricDiagnostics.probe` registers a task action. If that probe gets retired, this change becomes dead weight — harmless, just unnecessary. Flagging so it isn't read as load-bearing.

And it may well be retire-able. Every question the fingerprint probe was added to answer is answered and published in that comment:

- jar integrity — *"We hash-verified the android-all jar… sha1 + size match. The jar is intact."*
- bytecode-agent inventory — *"the agent set varies — it does not point to one culprit"*
- classload source — *"defined by its instrumenting classloader… not the stub android.jar"*

`TestFailureLogging` looks different to me and I'd argue for keeping it: the comment notes the primary NPE is *"normally swallowed unless full stack traces are forced"*, which is exactly what it does. Upstream is still open with no maintainer reply since 2 Jul, and the comment puts the rate at ~20 builds/week across 7 orgs — so it's still the only reason an affected customer sees `AndroidVersions.CURRENT is null` rather than the misleading `NoSuchFieldError` cascade. It's configuration-time only, no CC or IP hazard.

Either way that's a separate change, deliberately not bundled here.

### 2. `beforeProject` slightly reduces `TestFailureLogging` reach in the root project

The precedence fix has one consequence worth weighing explicitly. `configureEach` actions run in registration order, so moving to `beforeProject` means a project's own config is applied last and wins. That is the point for `robolectric.dependency.repo.url` — it keeps the documented escape hatch working — but it applies to `testLogging` too.

Net effect: a **root project** that sets its own `testLogging` now overrides our FULL-stack-trace setting. Previously ours won there. Subprojects already behaved this way (the old `allprojects`-from-root registration also ran before their build scripts), so this only shifts root-project behaviour.

In practice Android test tasks live in subprojects, so I expect this to be irrelevant — but if the FULL-stack-trace capture is the piece still earning its keep, it is a marginal loss of capture coverage and your call rather than mine.

### Also worth a follow-up ticket

Migrating `beforeProject`/`afterProject` to the isolated `gradle.lifecycle.*` API would be fully IP-clean. Not done here: those actions cannot capture `log()`, the `configureMirror`/`useMirror` `Action`s, or the `loggedReplacements` dedupe set, so it means rewriting the whole script plus a Gradle 8.8 floor. Per [gradle/gradle#30075](https://github.com/gradle/gradle/issues/30075) the current hooks are "not inherently incompatible with IP", so there's no correctness pressure to do it now.

PR #443, opened 2026-07-27
[ACI-5241] Auth and shell improvements, bugfixes

zsolt-marta-bitrise:
[ACI-5241](https://bitrise.atlassian.net/browse/ACI-5241)

The xcodebuild health-check gate that used to live here is now **#444**, stacked on this branch. This PR is the activation, auth and doctor work.

## What this fixes

- `activate --interactive` signs you in through the browser instead of making you create a PAT by hand.
- The browser login now survives a machine where localhost isn't reachable (RDE), by accepting a pasted callback URL.
- Interactive setup starts the proxies the selected tools need, so they aren't per-terminal-session.
- `activate` no longer destroys a browser login by rewriting the stored credential without its refresh token.
- `auth logout` finds the browser login again (it was the same destroyed credential).
- `auth logout` and `auth clear` now mean clearly different things, and each only clears what it says.
- Bazel uses the credential helper for OAuth too, instead of a token literal that goes stale.
- `auth set`, `auth clear` and `auth logout` work on a host with no OS keychain (headless Linux) instead of failing.
- `doctor --fix` repairs everything unprompted like `lint --fix`; `--fix --interactive` asks first.
- `esc` aborts every interactive form, and the abort is advertised in the help.
- The workspace picker no longer lags or drops keystrokes on accounts with many workspaces.
- `activate --interactive` no longer kills its own process on a machine that has activated before.
- Signing in moves credentials out of the config file into the keychain instead of leaving a copy behind.
- A rejected token stops burning the retry budget on every upload, download and FindMissingBlobs call.
- The benchmark phase is no longer written into your shell rc files, where it disabled the cache permanently.
- The enrichment retrier sweeps orphans at startup instead of waiting out a full interval.
- `activate bazel` / `activate gradle` name the CLI so `$PATH` resolves it, instead of baking in a path that breaks when the binary moves or was temporary.
- The Bazel credential helper now names `doctor --fix --interactive` when it finds no credentials, instead of only blaming an env var.
- A sign-in slow enough to take over standard input now stops and says how to continue, instead of letting the next prompt drop keystrokes.
- The interactive terminal flows live in one package now, and the OAuth package no longer owns stdin plumbing.

## Summary

Local-dev-setup gaps found on the 07.27 discovery meeting, plus the bugs that showed up while testing the fix on a real machine. Setting up the cache locally still required steps the CLI could do itself — creating a personal access token by hand, starting the proxies per terminal session — and several of the auth paths around it were quietly lossy: a browser login could be destroyed by the next `activate`, `auth logout` couldn't find it, and Bazel ignored it entirely.

`activate --interactive` now signs you in through the browser and starts the proxies your selected tools need, the OAuth flow survives a host where the browser can't reach the CLI's loopback listener (RDE), a browser login survives activation and works for every build tool, and `doctor --fix` behaves like `lint --fix` with an opt-in picker.

## Changes

**1. Interactive setup launches the OAuth login**

Auth is resolved *before* the huh form — the sign-in prints logs, opens a browser and reads stdin, none of which composes with a full-screen form. When nothing is stored and the env vars aren't set, the wizard announces the sign-in and waits for Enter (no browser opening out of nowhere); `s`/`skip`/`n` or EOF falls back to the manual token prompt, as does a failed login. A stored OAuth login is now also refreshed, which the wizard previously didn't do — an expired one was used as-is. When the sign-in is skipped because credentials already exist, the first screen says where they came from, so it's clear which identity is about to be activated.

**2. Callback-URL paste fallback for RDE sessions**

`oauth.Config.PasteReader` (set to stdin only when it's a terminal) races the loopback callback against lines pasted into the terminal. The hint is printed upfront, next to the authorize URL. `parsePastedCallback` accepts a full URL, a bare `?query`, or just the params; an unusable paste re-prompts, while a provider denial or a state mismatch ends the wait. Validation is shared with the HTTP handler via `parseCallbackParams` so the two paths can't drift.

**3. Interactive setup brings the daemon up**

After activation, only the services the selected tools need get registered and started — Xcode → `xcelerate-proxy`, ccache → `ccache-helper`; Gradle and Bazel talk to the cache directly and get none. Behind a confirm step, since it writes a LaunchAgent / systemd unit. Failures warn with the manual `daemon install` fallback rather than failing an activation that already succeeded.

Tool selection runs as its own form so the daemon question is only *built* when the selection needs a proxy — huh's accessible mode (`TERM=dumb`) walks every group and ignores `WithHideFunc`, so hiding it would still have prompted a Gradle-only user. Pinned by `TestHuhAccessibleModeIgnoresGroupHideFunc`.

**4. `activate` no longer destroys a browser login, and Bazel uses it**

`PersistActivateCreds` rebuilt the stored credential from just the token and workspace, so activating after a browser sign-in wrote back a record with the refresh token, JWT, expiries and username dropped. That's why `auth logout` reported nothing to log out: the OAuth-managed record it looks for no longer existed. Activation now merges into the existing record when the token still matches, and replaces it only when it doesn't.

Two related fixes in the same area:

- `oauth.Clear()` wiped both credential backends, contradicting the message it printed. It now clears only the backend the login was found in (`ClearFrom(source)`), which is also what makes `auth clear` (drop everything) and `auth logout` (end the browser session) mean different things.
- Loading now prefers an OAuth-managed credential over a manual one when both backends hold something, instead of taking whichever backend was checked first.
- A keychain that refuses to store falls back to the multiplatform config file rather than failing the login outright.

Local Bazel activation wrote a literal `Bearer <token>` into `~/.bazelrc`, which cannot work for OAuth: the PAT is short-lived and the file would hold a stale one. Local dev now routes both PAT and OAuth through `--credential_helper=*.services.bitrise.io=<CLI>`, so Bazel asks the CLI per request and gets a refreshed token, and no credential lands on disk. CI is unchanged and still gets the literal header — `bazelrc.gotemplate` gates the helper on `CLIPath && !CIProvider`, deliberately, to avoid spawning the CLI per RPC.

**5. `doctor --fix` behaves like `lint --fix`; the picker is `--interactive`**

`--fix` repairs everything it can without asking (skipping the auth fix, which needs a terminal, with a note). `--fix --interactive` prints the recap first, then a picker with errors preselected and warnings opt-in, and offers the OAuth sign-in for an auth problem by reusing the wizard's sign-in code rather than a second implementation. `--interactive` implies `--fix`, and fails on `--json` or without a TTY. Logs that suggest a repair name `doctor --fix --interactive`, matching what `activate` advertises.

**6. Escape aborts every interactive form**

Abort was `ctrl+c` only, unadvertised, and came out as a wrapped error plus a cobra usage dump. A shared keymap (`internal/tui`) binds `esc` alongside `ctrl+c` on every form and shows it in the help, including the manual PAT fallback prompt; huh's abort becomes `ErrAborted`, so the wizard, `auth login` and the doctor picker all report a plain cancellation. `esc` is safe to take because its other huh bindings (set/clear filter) are only enabled while filtering is on, which no field here turns on — the trade-off is that enabling type-to-filter on the picker later would need `esc` back.

**7. Workspace picker performance**

The picker lagged between items on accounts with many workspaces. huh applies its default height only to dynamic `OptionsFunc` fields, so a static `Options()` list renders in full — measured at 81 workspaces, that's an **82-line, 6.1 KB frame repainted per keypress**; `Height(12)` brings it to 12 lines and 923 B. (The residual per-keypress cost is linear in option count inside huh, ~3 ms at 81 — fine here, but it's why a much larger list would still drag.)

Separately, the picker dropped keystrokes and sometimes hung: the OAuth paste-reader goroutine was still blocked on stdin when the form started, so the two competed for input. The reader is now only armed after a grace period, so a browser that reaches the loopback listener never touches stdin at all.

When the grace *does* elapse — a slow SSO + MFA sign-in — the reader cannot be unarmed: Go treats character devices as non-pollable, so `SetReadDeadline` is rejected for any tty, and (measured under a pty) opening `/dev/tty` with `O_NONBLOCK` doesn't change that. So the paster reports that stdin is spoken for, `loginAndStore` returns `ErrStdinUnusable` rather than showing the picker, and each caller offers a way through without a prompt: `auth login --workspace <slug>` (with the workspaces listed, since it can still ask the API), the new `activate --interactive --workspace <slug>`, or simply rerunning the wizard — the credential is already saved, so the second run skips the sign-in and has a clean stdin. Raised in review; raising the grace instead would only move the cliff.

**8. `activate` no longer terminates its own process**

`makeSureCLIIsNotRunning` killed every running CLI process including the one doing the activating, so `activate --interactive` died mid-run on any machine that had activated before. It now skips its own PID, and `copyCLIToXcelerateBinDir` short-circuits when the binary is already in place instead of copying a file onto itself.

**9. Keychain migration is a move, not a copy**

Signing in scrapes credentials out of the multiplatform config file into the keychain. It left the file copy behind, so a token stayed on disk after a migration that was supposed to remove it. Verified on an RDE before and after: file only → keychain only.

**10. A rejected token no longer burns the retry budget**

An `Unauthenticated` upload was retried four times over ~3.3 s. A 401 is not transient, so the KV writer now aborts the retry loop on it (1 attempt, ~340 ms) and the error carries a sentinel the callers match on.

The read paths had the same problem in two places. `download` matched `Unauthenticated` with a bare `status.FromError`, which misses the rejection when the stream surfaces it wrapped, and `findMissing` had its abort flags **inverted** — it retried a rejected token three times and aborted on the transient errors the retry wrapper existed for. Both now go through the same `isUnauthenticated` check.

**11. The auth commands work without an OS keychain**

A Linux host with no secret-service on the session bus has no keychain — normal for headless machines, containers and most CI. `auth set` failed outright there, `auth logout` failed while reading, and `auth clear` aborted on the keychain *before* reaching the config file, so there was no way to remove the credentials the machine actually had.

Keychain errors that mean "there is no keychain here" are now classified as `keychain.ErrUnavailable`, matched on the **dbus error name** (`org.freedesktop.DBus.Error.ServiceUnknown`, `Spawn.*`, `NoServer`) rather than its human-facing message, plus `ErrUnsupportedPlatform` and the one case godbus reports as a plain error (no session bus at all). Reads fall through to the next backend, `auth set` falls back to the config file, `auth clear` skips the keychain and carries on, and `doctor`'s keychain smoke test warns instead of erroring — a host keeping its credentials in the config file is a working setup, not a broken one.

`auth status` still shows the distinction, since it reads the keychain directly rather than through the store.

**12. Generated configs no longer point at a temporary CLI path**

`activate bazel` and `activate gradle` baked `os.Executable()` into `~/.bazelrc` and the Gradle init script. Run through `go run .`, that is a path under `$TMPDIR/go-build*/` which Go deletes when the command exits — so `--credential_helper=*.services.bitrise.io=<path>` and the init script's `commandLine("<path>", "auth", "token")` both pointed at a binary that was already gone. Same for a release tarball unpacked in `/tmp`. Reported from a local Bazel run.

Both consumers can name the CLI without a path: the Gradle init script runs a bare command, and Bazel looks up a credential-helper value containing no path separators in `$PATH` (verified against Bazel itself, below). So `internal/clibin` resolves to the bare `bitrise-build-cache` whenever that name resolves — **not only when the running binary is on a transient path**, which is where I first drew the line and which still pinned an absolute path for the ordinary case of an installed CLI. Those paths go stale the moment an upgrade moves the binary, which is the whole reason to prefer `$PATH`.

The real path is the fallback for a host where nothing answers to that name, and when neither is available the caller decides: Bazel keeps the literal header (a token on disk still authenticates, an unspawnable helper fails every build), Gradle keeps the bare name plus a warning. The decision is a pure function, `resolvePath(exe, onPATH)`, so all four combinations are covered rather than just the ones a test binary's own path can reach. The bare-name default is applied while building the template inventory, not in the template: `bazelrc.gotemplate` still requires a non-empty `CLIPath`, so an inventory built without one can't produce `--credential_helper=<scope>=` with nothing after the `=`.

`internal/clibin` also absorbs the transient-path copy that `daemon install` needs — a LaunchAgent has no useful `$PATH`, so a supervisor really does need an absolute path. It used to live in `internal/daemon`, unreachable from the activate commands.

**13. The Bazel credential helper says how to fix missing auth**

Bazel spawns the helper per RPC and surfaces only its stderr, so that one line is all an unauthenticated build gives you. It used to be the resolver's raw error — `BITRISE_BUILD_CACHE_AUTH_TOKEN or BITRISEIO_BITRISE_SERVICES_ACCESS_TOKEN environment variable not set` — which is misleading here: the keychain and the config file were checked too, and setting an env var is the *least* likely fix for a local build. It now states all three sources were empty and names `bitrise-build-cache doctor --fix --interactive`.

**14. Two messages that could name the wrong backend**

`doctor --fix --interactive` reported `saved credentials to OS keychain` unconditionally, but both prompts behind it can land in the config file (on CI, or when the keychain refuses the write). The location is gone from that line — its signature can't tell, and `auth status` is the source of truth for where a credential lives.

The wizard made the same assumption: a fresh login was recorded as `AuthSourceKeychain` regardless of where `loginAndStore` actually put it. It now reports the `store.Kind` it used, so a keychain-less host describes the config file and writes a later display-name change there, instead of warning about a keychain that isn't present. Both raised in review.

**15. Benchmark phase no longer written into shell rc files**

`ApplyBenchmarkPhase` exported `BITRISE_BUILD_CACHE_BENCHMARK_PHASE_*` into `~/.bashrc` and `~/.zshrc`. The phase is per-build, but an rc entry outlives the build — and `GetBenchmarkPhase` short-circuits on the env var *before* calling the API, so a single baseline result pinned the phase permanently. Baseline disables the cache, so local Gradle builds silently stopped using it. Persistent self-hosted runners with a shared home were affected identically.

Nothing on CI read it. Bitrise and GHA both run steps in non-interactive shells, which don't source `.bashrc`/`.zshrc` (verified: `bash -c`, `bash -e -c` and `bash -l -c` all see nothing; only `bash -i` does). Propagation stays `envman` / `GITHUB_ENV`, plus the phase file — which is what React Native's activator and `LogBenchmarkSummary` already read *exclusively*, and what the Xcode wrapper falls back to when a GHA step activates and builds in one shell. The `BITRISE_BUILD_CACHE_BENCHMARK_PHASE_XCODE` pin in `bitrise.yml` works through the env-var short-circuit and is untouched.

The rc export was never part of the design — `bc03e37` describes only the env var and the file. It came along with `EnvExporter`, whose shell-rc method exists for the xcelerate `PATH` block (a legitimate use: `PATH` *must* persist into interactive shells). Both benchmark appliers now take a narrower interface with only `Export`, so the method isn't reachable.

**16. Enrichment retrier sweeps at startup**

`Retrier.Run` only swept on its ticker, so orphans left by a previous process sat for a full interval (5 min default) while the doctor reported them as pending. It now sweeps once before the first tick.

`TestRetrier_StartupOrphanSweep_RemovesOldUntouched` asserted this but could never have passed: it built records with `Attempts=0`, a shape no code path writes (`Enricher` always sets `Attempts=1` + `FirstAttempt`) and which `Sweep` skips *before* the age check. Reshaped to what the enricher actually stores and renamed to what it verifies. Note `Sweep` still ignores `Attempts==0` records outright, so any written by an older CLI would leak — flagged, not changed here.

**Supporting refactors** (first commit, no behavior change)

- `internal/doctor` exports its result API (`EffectiveState`, `EffectiveOverall`, `ItemDisplay`, `Partition`). `cmd/doctor` delegates to it instead of keeping its own copies.
- The supervisor-registration logic (transient-binary copy, shadowed-binary warning) moved from `cmd/daemon/install.go` into `internal/daemon`, since `cmd/common` can't import `cmd/daemon` (import cycle).
- `internal/tui` holds the shared form chrome (abort keymap, select height/width, `RunForm`) so `internal/authprompt` and `cmd/doctor` can use it without importing each other.
- **The interactive terminal flows moved to `cmd/common/interactive`.** `cmd/common` had accumulated fifteen files for the wizard, the browser login, the form chrome and the prompts, interleaved with the rest of the command layer; they now sit in one package and lose the `interactive_`/`activate_interactive_` prefixes that were standing in for a package name. `hydrateStoredAuth` stays behind in `cmd/common` (the root command calls it, and it needs nothing interactive), which is what keeps the dependency one-way: `interactive` → `common`, never back.
- **`internal/oauth` no longer owns stdin.** It takes a `CallbackFallback` — a hook supplying the authorization code when the browser can't reach the loopback listener — and keeps only the race and the state check. The terminal half (the grace period, the read deadline, the warning when it can't be unarmed) lives with the caller. `ParsePastedCallback` is exported so a fallback can't skip the state check.
- **The interactive terminal flows moved to `cmd/common/interactive`.** `cmd/common` had accumulated fifteen files for the wizard, the browser login, the form chrome and the prompts, interleaved with the rest of the command layer; they now sit in one package and lose the `interactive_`/`activate_interactive_` prefixes that were standing in for a package name. `hydrateStoredAuth` stays behind in `cmd/common` (the root command calls it, and it needs nothing interactive) which is what keeps the dependency one-way: `interactive` → `common`, never back.
- **`internal/oauth` no longer owns stdin.** It takes a `CallbackFallback` — a hook that supplies the authorization code when the browser can't reach the loopback listener — and keeps only the race and the state check. The terminal half (the grace period before stdin is armed, the read deadline that unblocks it, the warning when it can't be) lives with the caller. `ParsePastedCallback` is exported so a fallback cannot skip the state check.
- `log-dirs` now probes only the log dirs of activated tools (`Doctor.stateDirCandidates`). It previously checked both the xcelerate and the ccache dir unconditionally, so an Xcode-only setup was warned about a missing ccache log dir. Every other tool-specific check already skips when its tool isn't activated.

## Testing

```
make lint                                  # 0 issues
go test -tags unit -race ./...             # all packages pass
```

### e2e

- `local_e2e_scenarios.sh` asserts `log-dirs` doesn't probe the ccache dir when ccache isn't activated (the script's cleanup removes `~/.bitrise/cache/ccache`, so this is deterministic).
- `local_e2e_scenarios.sh` asserts activation leaves no `BENCHMARK_PHASE` in any shell rc file.
- `local_e2e_scenarios_macos.sh` drives the wizard in accessible mode selecting Xcode and asserts it reached launchd with the xcelerate proxy *and only that* — the mac VMs have a usable launchd user session, so this needs no RDE.
- `rde_smoke_scenarios.sh` SCENARIO D drives the callback-URL paste fallback under a pty: RDE is the environment the path exists for (the browser runs on the user's laptop, not the VM), and the printed authorize URL supplies the `state`, so both the accept and reject paths are exercised without an IdP round-trip. Note the `rde-*-local-dx-smoke` workflows are currently commented out of the `features-e2e` pipeline, so this won't execute until RDE provisioning is fixed.

### Unit

New tests:

- `internal/oauth/callback_fallback_test.go` — paste parsing table (full URL, bare query, params, quoted/whitespaced, denial, state mismatch, unusable input), plus `Login` tests driving the paste path end-to-end against the existing OAuth mock, including "garbage line then the real URL" and one that fails without the stdin-race fix.
- `internal/auth/store/persist_merge_test.go` — activation merging into an existing OAuth record, and replacing it when the token differs.
- `cmd/common/interactive/wizard_auth_test.go` — no credentials → signs in; env vars → no sign-in; declined/failed sign-in → manual prompt; stored OAuth login refreshed; unrefreshable login → signs in again; manual keychain credential used as-is; keychain re-read after sign-in (so a later display-name write can't drop the OAuth tokens); the confirm prompt's input handling; the credential-source note; and that no browser is launched when nothing can confirm it (accessible mode owns stdin), which otherwise blocked a scripted run for the full 5-minute callback timeout.
- `cmd/common/interactive/daemon_test.go` — tool → service mapping, plus a guard pinning that huh's accessible mode ignores group hide funcs.
- `cmd/doctor/` — `--interactive` implying `--fix`, the `--json` and no-TTY conflicts, and the picker preselecting errors but not warnings.
- `internal/tui/` — the abort keymap binding `esc`, and `ErrAborted` translation.
- `internal/build_cache/kv/` — an `Unauthenticated` close aborts the retry loop instead of retrying, and `isUnauthenticated` recognises both the bare status and a wrapped one.
- `internal/auth/keychain/unavailable_test.go` — the dbus `ServiceUnknown` error (captured from the headless VM) classifies as unavailable while a *locked* keychain does not, and Load/Save/Clear all report it while keeping the original cause.
- `cmd/auth/auth_test.go` — an unavailable keychain doesn't stop `auth clear` reaching the file store, and a real failure in one backend still doesn't skip the other.
- `internal/oauth/callback_fallback_test.go` — the flow takes a code from the fallback, can't have the state check bypassed by one, and **waits for a cancelled fallback to release stdin** before returning (this one fails deterministically without the wait).
- `internal/config/gradle/benchmark_test.go` — `ApplyBenchmarkPhase` creates no rc file, driven through the real `envexport` against a temp `HOME`.
- `internal/doctor/report_test.go` — `EffectiveOverall`, `Partition`, and log-dir candidates scoped per activated tool.
- `internal/clibin/clibin_test.go` — transient-path classification (including the exact `go run` path shape), the copy landing on the stable path, and a table over `resolvePath` covering every (stable/transient × on-PATH/not) combination, plus two end-to-end cases through the real process state.
- `internal/config/bazel/bazel_config_generate_test.go` — a bare `CLIPath` still emits the helper line and no `authorization` header; the empty-`CLIPath` and CI cases still get the literal header.
- `internal/bazelcredhelper/helper_test.go` — the no-credentials error names the doctor command, doesn't blame env vars alone, emits no header, and stays on one line. It calls `keyring.MockInit()`: HOME isolation alone isn't enough, since the resolver reads the real OS keychain and the test would otherwise pass for the wrong reason on a machine with stored credentials.
- `cmd/common/interactive/callback_paste_test.go` — the paster returns the code from a pasted URL, nudges on a mis-paste and keeps reading, treats a state mismatch as final, never touches the reader before the grace elapses, reports stdin unusable when it can't stop a terminal-like reader, and omits the `--workspace` advice for a caller that has no such flag.

Each commit was verified to build standalone.

Manually verified on an RDE macOS VM (the environment several of these bugs only reproduce on): the wizard's sign-in + daemon start, the paste fallback, `activate --interactive` on a previously-activated machine, `auth logout` after a browser login, the keychain migration leaving nothing on disk, and Bazel resolving a token through the credential helper.

The CLI-path fix was verified against the reported repro: `go run . activate bazel` with a throwaway `HOME`, then running the pinned helper *after* `go run` exited — it answers with a real `Bearer` token instead of failing on a missing binary. `activate gradle` writes `commandLine("bitrise-build-cache", ...)` and warns when that isn't on `$PATH`.

Verified on the Linux RDE, which is where the absolute-path report came from: `activate bazel` from the installed CLI writes `--credential_helper=*.services.bitrise.io=bitrise-build-cache`. The stable-path-not-on-`$PATH` case was checked separately (outside `/tmp`, since the transient classifier matches that) and correctly pins the absolute path.

The `$PATH` lookup itself was verified with Bazel rather than from its docs — a real `bazel build` against the generated flag, with a logging shim named `bitrise-build-cache` on `$PATH`:

```
args=get in={"uri":"https://ci-cache.services.bitrise.io/build.bazel.remote.execution.v2.Capabilities"} out={"headers":{"authorization":["Bearer faketoken12345"]}}
```

Bazel spawned the bare name and accepted the CLI's response.

The keychain-less path was verified on the same headless **Linux** RDE, where it reproduces: before, `auth set` / `clear` / `logout` all exited 1; after, `set` falls back to the config file, `clear` skips the keychain and removes the file credential, `logout` finds the login in the file store, and `doctor` reports a warning rather than an error. The dbus error name in the classifier was read off that machine rather than guessed.

## Notes for reviewers

- **Test isolation, and why five packages were red.** The suite read credentials from HOME-resolved config and wrote into `~/.local/state` / `~/.bitrise`, so results depended on the developer's own machine. A per-package `TestMain` now points `HOME` at a throwaway dir. That alone fixes `cmd/bazel`, `cmd/gradle`, `internal/config/bazel` and `internal/config/gradle` (they expected a missing credential and found the developer's); `pkg/ccache` was finding one an earlier subtest left in the shared mock keyring, reset in place. Two further tests were time bombs against committed fixtures — the enrichment watcher's age gate drops entries once the fixture timestamps pass `HandledManifestMaxAge`, and the retrier sweep test dated its records from a hardcoded `now` while the code under test used the real clock; both had injectable clocks that simply weren't used. **Net: six failing packages down to zero.**
- **Worth a look:** the stdin handover in `cmd/common/interactive/callback_paste.go`. The deadline-based unarming only works for pollable readers (pipes, and the test doubles) — never for a real terminal — so the load-bearing parts are the grace period, which keeps the reachable-browser case from touching stdin at all, and `StdinUnusable()`, which stops every later prompt when the reader did get armed. `internal/oauth` waits for a cancelled fallback to release its reader before returning, bounded so a misbehaving one can't wedge the sign-in.
- **Order change in the wizard:** the sign-in now happens before the tool selection, not as a form group in the middle. That's what makes the browser + stdin interaction possible at all.
- `wizardStartingCreds` gained an injectable `resolve` parameter. Production `ResolveAuthConfig` reads the real keychain and on-disk config, which is exactly what made the existing tests in this repo machine-dependent; the new tests don't repeat that mistake.
- **Found via the rc fix:** running `internal/config/gradle`'s tests on the old code wrote the benchmark block into the *developer's own* `~/.bashrc` and `~/.zshrc`, because `Test_activateGradleParams` calls `ApplyBenchmarkPhase` with a real `envexport` and a real `HOME`. The narrowed interface stops that too. Those tests still write a real `~/.local/state/xcelerate/benchmark/*.json`, which is harmless but poor hygiene — worth a follow-up to inject a temp `HOME` there, not done here to keep the diff scoped.
- **Not covered by an automated test:** the keychain-write failure fallback *end to end*. The classifier and the `auth clear` behaviour are unit-tested against the real dbus error, but making a developer machine's OS keychain actually refuse a write isn't something the suite can do without mocking the layer the fix is about — that half was verified by hand on the Linux RDE.
- **Left alone deliberately:** `download`'s `initiateGet` error path aborts the retry loop for *every* error, not just the unauthenticated one, so a transient failure at stream setup never retries. It's the same inversion as `findMissing`, but it isn't an auth-handling bug and changing it alters retry behaviour on genuinely broken setups — worth its own look.

[ACI-5241]: https://bitrise.atlassian.net/browse/ACI-5241?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cmd/common/interactive/callback_paste.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	if err := dr.SetReadDeadline(time.Now().Add(-time.Second)); err != nil {
+		// Character devices reject deadlines, so this is the terminal case.
+		p.warnUnreliable(err.Error())
+
+		return
+	}
+	<-done
+	_ = dr.SetReadDeadline(time.Time{})
```
ramil-bitrise, line 130: **This branch is the only one that ever runs in production, so the reader is never actually stopped.**

Go constructs `os.Stdin` as non-pollable, so `SetReadDeadline` returns `ErrNoDeadline` for *any* tty — not just macOS. Verified under a pty:

```
mode=Dcrw--w---- chardev=true SetReadDeadline err=file type does not support deadline
```

So `stop()` always falls into `warnUnreliable` and returns; the `<-done` wait and the deadline restore on 115-116 only execute in tests, where `io.Pipe` and `blockingReader` accept deadlines. `TestLogin_LoopbackWinCancelsTheFallbackAndWaits` tests the contract well, but nothing exercises a reader that behaves like the production one.

That leaves `pasteGrace = 15s` as the sole real protection against the picker keystroke theft this PR set out to fix: under 15s no reader is ever armed and all is well, but an SSO + MFA + first-time-consent sign-in routinely exceeds it — and then the reader is armed, unstoppable, and `pickWorkspace` runs immediately after, dropping keystrokes exactly as reported.

Roughly in order of preference:

- Thread the "stdin is compromised" state out of `Fallback` → `loginAndStore` and skip the interactive picker in that case (fail asking for `--workspace`, or let the still-live paste reader supply the answer).
- Open `/dev/tty` yourself with `O_NONBLOCK` for the paste reader instead of borrowing `os.Stdin`, so a deadline is honoured.
- Failing both, raise the grace — 15s is below the median real sign-in.

cmd/common/interactive/callback_paste.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+func (p callbackPaster) warnUnreliable(reason string) {
+	p.warnf("Could not stop reading standard input (%s).", reason)
+	p.warnf("Later prompts in this command may drop keystrokes — pass --workspace to skip the picker.")
```
ramil-bitrise: The remedy offered here doesn't exist in the path where this warning is most likely to fire: the wizard calls `loginAndStore(ctx, r.Logger, r.Envs, "", "")` and `activate --interactive` has no `--workspace` flag. `auth login` does, so the advice is right for that caller and unactionable for the other.

Either add the flag to `activate --interactive`, or branch the message on whether the caller can actually supply a workspace.

cmd/common/interactive/wizard_auth.go
```diff
@@ -0,0 +1,202 @@
...
+
+	creds, ok := r.signIn(ctx)
+	if !ok {
+		return auth
+	}
+
+	auth.Config = configcommon.CacheAuthConfig{AuthToken: creds.PAT, WorkspaceID: creds.WorkspaceID}
+	auth.Source = configcommon.AuthSourceKeychain
+	auth.Stored = loadWizardKeychain(r.Logger, r.Keychain)
+	auth.SignedInNow = true
```
ramil-bitrise: `Source` is hardcoded to `AuthSourceKeychain`, but `loginAndStore` picks its backend via `store.Select` and — thanks to `saveLoginWithFallback` — can legitimately land in the file store: on CI, or when the keychain refuses the write, i.e. the keychain-less host that change #11 exists to support.

On such a host the consequences compound: `loadWizardKeychain` on line 86 comes back empty, so `auth.Stored` is empty; `resolvedAuthNote` would describe the wrong backend; and a display-name edit later takes the `AuthSourceKeychain` branch of `persistWizardCredentialsTo`, calls `persistCredentials(kc, …)`, hits the missing keychain and warns — even though the credential is sitting in the config file and is perfectly writable.

Suggest returning the `store.Kind` from `loginAndStore` (it already computes it for its own log line) and mapping it to `AuthSourceKeychain` / `AuthSourceFile` here.

cmd/common/interactive/fix_auth.go
```diff
@@ -0,0 +1,56 @@
...
+			logger.Infof("Falling back to entering a token.")
+
+			return authprompt.PromptAndSave() //nolint:wrapcheck // caller reports it
+		}
+
+		if creds.PAT == "" {
+			return "", "", fmt.Errorf("sign-in returned no token")
+		}
+
+		return creds.WorkspaceID, creds.PAT, nil
```
ramil-bitrise: This hands `(workspaceID, PAT)` back to `AuthPromptFixer.Fix()`, which reports the outcome as a hardcoded string:

```go
return "saved credentials to OS keychain (workspace=" + workspaceID + ")", nil
```

That was accurate when `authprompt.PromptAndSave` was the only prompt. It no longer is: `loginAndStore` may have stored to the multiplatform config file (CI, or the keychain-refused fallback), and so may `authprompt.PromptAndSave` on line 40. So `doctor --fix --interactive` can tell the user the credential went to the keychain when it went to disk — which is exactly the distinction the rest of this PR works hard to keep straight.

Either widen the prompt signature to return the backend, or drop the location from the fixer's message and let `auth status` be the source of truth.

internal/config/bazel/activate_bazel_params.go
```diff
@@ -198,3 +200,14 @@ func (params ActivateBazelParams) rbeTemplateInventory(
...
+// credentialHelperPath keeps the helper reachable when the caller has no usable
+// absolute path: Bazel resolves a bare name through $PATH, which also survives a
+// CLI upgrade that moves the binary.
+func credentialHelperPath(cliPath string) string {
+	if cliPath != "" {
+		return cliPath
+	}
+
+	return paths.CLIBinaryName
+}
```
ramil-bitrise, line 222: This makes the inventory's `CLIPath` unconditionally non-empty, which defeats the template's own safety net. `bazelrc.gotemplate` still gates on `and .Common.CLIPath (not .Common.CIProvider)` and its comment still promises *"An empty CLIPath still falls back to the literal header rather than emitting a helper flag with no path"* — but nothing reaching the template through `commonTemplateInventory` can hit that branch any more. Only the hand-built inventories in `Test_Generate` do, so the tests that appear to cover the fallback are now covering a path production can't reach. Worth saying so in the comment at minimum.

Where that bites: `internal/clibin/clibin.go` `Resolve` returns `""` when `os.Executable()` fails (line 63) **without** calling `warnIfNotOnPATH` — that check only guards the transient-path branch on line 73. Combined, a host where the CLI isn't on `$PATH` silently gets a `~/.bazelrc` whose credential helper can never be spawned, so every Bazel build fails to authenticate. Previously that host got a working, if token-on-disk, config.

Two asks:

1. Move `warnIfNotOnPATH(logger)` so it covers both `return ""` paths in `Resolve`.
2. Consider keeping the literal-header fallback when `exec.LookPath` actually fails — the template gate is already written for it, so it's a matter of not overriding the empty `CLIPath` when the bare name provably won't resolve.

ramil-bitrise (commented):
Reviewed against the head branch (built it, ran the touched packages' unit tests — all green). The body of this PR is genuinely excellent, and the tests are unusually good at explaining why they exist; pinning huh's accessible-mode hide-func behaviour is exactly the right instinct.

Five things I'd want looked at, left as inline comments. The first two are the ones I'd block on — the paste-reader stop path can't take effect on a real terminal, so fix #7 currently rests entirely on the 15s grace. The other three are a few lines each.

Separately, and not blocking: 15 independent behaviour changes plus four refactors across 90 files makes bisecting a regression to any single fix painful, and the commits already separate cleanly (`internal/tui` + `internal/clibin` + `cmd/common/interactive`, the kv retry fix, and the keychain-availability work would each stand alone).

cmd/common/interactive/callback_paste.go
zsolt-marta-bitrise: Confirmed, and thanks — this was the important one. Fixed in a68658f.

I checked your preferred option (b) before building it, and **it doesn't work**: owning the fd changes nothing, because Go refuses to poll character devices regardless of how they were opened. Measured under a pty:

| reader | `SetReadDeadline` |
|---|---|
| `os.Stdin` | `file type does not support deadline` |
| `/dev/tty` blocking | `file type does not support deadline` |
| `/dev/tty` + `O_NONBLOCK` | `file type does not support deadline` |

So I went with (a). `callbackPaster` is now a pointer with `StdinUnusable()`, set wherever `stop()` gives up. `loginAndStore` checks it before `pickWorkspace` and returns `ErrStdinUnusable` instead of racing it, and each caller says how to continue without a prompt:

- `auth login` → prints the workspaces it can still list with the fresh PAT (the picker that would have shown them is what's unavailable) and the `--workspace <slug>` rerun.
- `activate --interactive` → the credential is already saved, so it stops and tells you to rerun; the second run resolves the stored credential, never signs in, and has a clean stdin.
- `doctor --fix --interactive` → propagates the sentinel rather than falling back to the token prompt, which would drop keystrokes too.

I did not raise the grace: with the state threaded out, exceeding 15s is now handled rather than merely unlikely, and a longer grace would delay the paste path for the RDE users it exists for.

New test `TestCallbackPaster_SlowSignInOnATerminalReaderReportsUnusableStdin` drives exactly your scenario — grace elapses, reader armed, `os.ErrNoDeadline` on stop — and asserts the reader really was read from and that the state surfaces.

cmd/common/interactive/callback_paste.go
zsolt-marta-bitrise: Right on both counts — fixed in a68658f, using both halves of your suggestion.

`callbackPaster` now takes a `WorkspaceFlag` from the caller, so the message only offers what that command actually accepts, and `activate --interactive` gained a real `--workspace <slug>` flag so it has something to offer. `doctor --fix --interactive` passes an empty flag name and gets "it will stop rather than drop keystrokes" instead.

Pinned by `TestCallbackPaster_OmitsTheFlagAdviceWhenTheCallerHasNoFlag`, which asserts `--workspace` is absent for a caller that has no such flag.

cmd/common/interactive/wizard_auth.go
zsolt-marta-bitrise: Good catch — the compounding you describe is real. Fixed in a68658f, the way you suggested.

`loginAndStore` now returns a `loginOutcome` carrying the `store.Kind` it used, and `wizardAuth` records it. `Source` is derived (`authSourceForKind`), `Stored` is re-read from that same backend, and `persistWizardCredentialsTo` handles `SignedInNow` before the source switch: the credential is already persisted, so all that's left is a changed display name, written to the backend the login actually used.

That removes the spurious "Could not save credentials to the OS keychain" on a keychain-less host, which previously came from the `AuthSourceFile` branch trying to migrate a credential that was deliberately in the file.

cmd/common/interactive/fix_auth.go
zsolt-marta-bitrise: Confirmed — fixed in a68658f with your second option.

I dropped the backend from the message rather than widening `Prompt func() (workspaceID, authToken string, err error)`: that signature is an exported field on `Doctor`, and it has two implementations that can each land in either backend, so plumbing the location through both to render one line isn't worth the API change. It now reads `saved credentials (workspace=X)`, with a comment saying `auth status` is the source of truth for where.

internal/config/bazel/activate_bazel_params.go
zsolt-marta-bitrise: Both taken, in a68658f — the second one is the more important of the two and I'd missed the consequence.

1. `warnIfNotOnPATH` now covers both `return ""` paths in `Resolve`.
2. `credentialHelperPath` only defaults to the bare name when `clibin.OnPATH()` says it resolves; otherwise it stays empty and the template's literal-header branch takes over. So a host without the CLI on `$PATH` gets a working config again instead of a helper that can never be spawned — a token on disk still authenticates, an unspawnable helper fails every build.

That also makes the fallback branch reachable from production again, so the `Test_Generate` cases covering it are meaningful rather than theoretical. The template comment now states when `CLIPath` is empty and why that path exists, instead of just asserting it.

internal/oauth/store.go
```diff
@@ -62,7 +62,15 @@ func LoadWithSource() (Credentials, store.Store, error) {
...
 }
 
+// loadFrom prefers an OAuth-managed credential over a manual one wherever it
+// lives: a plain `auth set` PAT in an earlier backend would otherwise hide a
+// login stored in a later one, so logout and refresh would both miss it.
 func loadFrom(backends ...store.Store) (Credentials, store.Store, error) {
+	var (
+		firstCreds Credentials
+		firstStore store.Store
+	)
```
ramil-bitrise, line 72: **Not a change request for this PR** — this change is correct and the comment above it explains exactly why it's needed. Flagging the pattern it lands in, for a future ticket.

This makes the *fourth* credential-precedence order in the codebase, each living in a different package with its own rules:

| Entry point | Order |
|---|---|
| `configcommon.ResolveAuthConfig` | env → JWT → keychain → file → multiplatform |
| `oauth.loadFrom` (here) | keychain → file, but prefers OAuth-managed |
| `interactive.wizardStartingCreds` | keychain-first, *then* delegates to `ResolveAuthConfig` |
| `store.SelectAuto` | CI→file, local→keychain (write policy, same knowledge) |

Around them: four types for one concept (`keychain.Credentials`, `configcommon.CacheAuthConfig`, `oauth.Credentials`, `multiplatformconfig.Credentials`) with `fromKeychain`/`toKeychain` conversions; three interfaces that are all just `Load()` (`store.Store`, `configcommon.AuthLoader`, `interactive.keychainStore`); and two provenance enums that don't line up — `configcommon.AuthSource` (6 values, separates File from Multiplatform) vs `store.Kind` (2 values, collapses them).

That last mismatch is the interesting one, because it's *why* two of my other comments exist rather than being independent bugs:

- `cmd/common/interactive/wizard_auth.go:85` hardcodes `Source = AuthSourceKeychain`
- `internal/doctor/fix_auth_prompt.go:27` hardcodes `"saved credentials to OS keychain"`

In both cases the write path *knew* which backend accepted the credential and there was no type able to carry that fact back to the caller. `store.SaveOutcome.Kind` — added in this PR — is the start of exactly that type; it just stops at the storage boundary instead of reaching the wizard and the doctor.

Shape worth aiming at, whenever this comes up next: one `auth` facade owning `Resolve(envs) (Credential, Origin, error)`, where `Origin` carries backend *and* provenance in a single type and `Credential` is one struct. Everything downstream — doctor, wizard, `auth status`, the Bazel credential helper, Gradle — consumes that instead of re-deriving it. The tell that it's worth doing: `wizardStartingCreds` exists only because `ResolveAuthConfig`'s order was wrong for one caller, and it now takes an injectable `resolve` parameter so tests can avoid the real machine. Both are workarounds for resolution not being a first-class, injectable thing.

Happy to write this up as a ticket if useful — just didn't want it to disappear into a review thread.

ramil-bitrise (commented):
One architectural observation, split out from my earlier review. **Explicitly not for this PR** — no change requested here, and I wouldn't want this refactor attempted inside a 90-file changeset. Filing it as context because this PR is where the cost became legible, and two of my earlier inline comments are downstream symptoms of it.

ramil-bitrise (approved):

internal/oauth/store.go
zsolt-marta-bitrise: Ticketed as **[ACI-5274](https://bitrise.atlassian.net/browse/ACI-5274)** under *ACI - Technical Improvements* (ACI-310), with your inventory as the body — nothing here should live only in a review thread. I searched ACI first; nothing existed (ACI-4897 `auth check` and ACI-5269 are adjacent but neither covers this).

Two things to fold in, both found after you wrote this, both instances of the pattern you described rather than new problems.

**The fifth resolution path arrived, and it's a second copy of an existing algorithm.** #450 adds `bazelcredhelper.newResolver`, which is step-for-step `configcommon.ExpiryAwareResolver.Get()`:

| step | `ExpiryAwareResolver` (xcelerate proxy) | `bazelcredhelper` (#450) |
|---|---|---|
| 1 | `resolveFn(envs)` → cfg, source | `ResolveAuthConfig(envs)` → cfg, source |
| 2 | resolve error → warn, serve cfg | resolve error → return error |
| 3 | `!source.StoreManaged()` → serve cfg | `!source.StoreManaged()` → serve cfg |
| 4 | `refreshFn(ctx)` | `ensureFresh(ctx)` |
| 5 | refresh error → debug, serve stale | refresh error → warn+cooldown, serve stale **+ expiry hint** |
| 6 | `CacheAuthConfig{pat, wsid}` | `Credential{pat, expiry}` |

Same control flow; different return type and failure policy.

**And the reason it had to be a second copy is the `RefreshFunc` seam.** `internal/oauth` already imports `internal/config/common` (via `internal/auth/store`), so `config/common` can never import `oauth` — hence `RefreshFunc` as a function-typed dodge. It flattens `oauth.Credentials` to `(pat, workspaceID)`, **discarding `PATExpiry`** — which is exactly the field #450 needs for the EngFlow spec's `expires`. So the reuse wasn't available: it was type-erased away by the cycle workaround.

That sharpens where the facade has to live. Not in `config/common` — it needs a small package *above* both (`internal/auth/live`, say) that can import `oauth` freely and return token + workspace + **expiry** in one type. `ExpiryAwareResolver` then deletes, along with the `//nolint:containedctx` on it (#450's per-call `ctx` signature is the better shape), and the proxy gains a real PAT expiry instead of its fixed TTL.

**Also worth recording against your `AuthSource` vs `store.Kind` point**, since it turned out to be load-bearing three more times since:

- `ExpiryAwareResolver.Get()` and `DescribeResolvedWith` both gated on `source == AuthSourceKeychain`, so **file-stored logins were skipped entirely** — a real break on keychain-less hosts, fixed in #450.
- `mergeActivateCreds` compared the resolved token with the stored one to decide "same credential". A login's PAT is short-lived and refreshes, so they routinely differ, and an ordinary `activate` discarded the refresh token. Reproduced on an RDE, fixed in #449.
- `ReadConfig` (the invocation PUT's credential) never consults env vars while `ResolveAuthConfig` prefers them, so the two disagree *at the same instant*. #444 papered over it by pinning a credential into the doctor's backend probe; that pin is gone now, so the divergence is documented and still live. It's in the ticket's DoD.

Your closing line reads differently in hindsight: *"both are workarounds for resolution not being a first-class, injectable thing."* Four auth fixes across #443/#449/#450 in two days, every one a symptom of that. The ticket argues for doing it before a sixth path shows up.

PR #450, opened 2026-08-04
fix: ACI-5269 refresh the auth token in the Bazel credential helper

viktorbenei:
[ACI-5269](https://bitrise.atlassian.net/browse/ACI-5269)

## TLDR

- The Bazel credential helper refreshes the token now instead of serving whatever is on disk forever, and returns the spec's `expires` so Bazel re-consults it before it goes stale.
- Logins stored in the config file — any host without an OS keychain — are refreshable too; the resolver and `auth status` both gated on the keychain.
- `EnsureFresh` serialises the refresh across processes, so the helpers Bazel spawns per RPC can't each spend the same rotated refresh token and break the login.
- Cross-process locking is `github.com/gofrs/flock`; `internal/filelock` and `internal/proxypid` are both deleted.
- The locking commits (`9e98bad` onwards) can be split onto their own branch if you'd rather review them apart from the fix — say so.

## Why the lock is a kernel lock, and where the old packages' pieces went

The first cut of this PR hand-rolled the lock: an `O_EXCL` marker holding the owner's pid, plus liveness probing, a TTL, break-open and steal logic to work out whether a holder had died. Two separate double-hold bugs turned up in that logic in a day — one from a flaky test, one from review — and each fix grew it further.

`flock(2)` locks are released by the kernel when the holder exits or is killed, which deletes the entire problem: no pid probing, no TTL, no stealing, no guard files. So `internal/filelock` went, and `internal/proxypid` — which had been folded into it — went with it.

Two pieces have no flock equivalent, and they moved to the callers that wanted them because those callers wanted different things:

- **`cmd/xcode/proxy_lock.go`** — the proxy pid file is now purely an advertisement, for `stop-proxy` and the xcodebuild wrapper to read. Exclusion is the kernel lock, and liveness is "can the lock be taken". `withProxySingleton` is the only way to take it, so contention always means *log and carry on* rather than a caller inventing its own policy.
- **`internal/bazelcredhelper/cooldown.go`** — the stale-credential warning's rate limiter. flock gives no TTL, so the mtime marker stays; the lock just stops N helpers all claiming the same window and printing N copies.

Two invariants the implementation now depends on, commented at both call sites: **the lock file is never unlinked** (a process holding the old inode and one locking a newly created file would both own it), and the pid is written **only after** the lock is held (a reader catching the truncate would otherwise conclude nothing is running).

`gofrs/flock v0.13.0` declares `go 1.24.0`, so it doesn't move our toolchain floor, and its only real dependency is `golang.org/x/sys`, already required.

## Testing

- On a keychain-less Linux RDE with a real `auth login`: both refresh paths work against live WorkOS/Bitrise — the JWT exchange rotates the PAT, and forcing the full grant rotates PAT, JWT and refresh token, all persisted. Against the pre-fix base, same seed: stale token, no `expires`, and **zero** requests to the IdP.
- `bazel build` end to end with `--credential_helper` — succeeds, BES upload clean, and the generated bazelrc contains no literal `authorization` header, so the token BES accepted could only have come from the helper.
- Multi-process integration tests (`internal/oauth`, `cmd/xcode`) re-exec the test binary so real processes contend. The IdP stub behaves like WorkOS — one use per refresh token, reuse rejected — so they assert the login survives rather than that holders didn't overlap. Green on macOS and Linux.
- Every one of those tests was checked to **fail** against its regression. That caught three that passed vacuously, twice because the code path short-circuited before the assertion could bite.
- `make lint` clean, `make test-unit` clean, lock packages re-run under `-race` for flakiness.

[ACI-5269]: https://bitrise.atlassian.net/browse/ACI-5269?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Badlazzor (approved):

zsolt-marta-bitrise (commented):
Reviewed against the PR checklist. I'd like this in v3.4.2 — it fixes a break #443 introduced (routing local Bazel through the credential helper) and closes the keychain-less hole #449 only half-closed. I'm holding the release tag for it.

**One request before merge: the comment trims below.** Everything else here is non-blocking — two questions you may have already weighed, and context from my side.

## Please trim before merging

The same three facts are each stated several times:

1. **"stdout is the protocol channel" appears three times.** `helper.go` says it in the type comment *and* again immediately below ("Nothing on this path may write to stdout — it is the protocol channel"), then `resolver.go` a third time. The `resolver.go` one earns its place — it explains why `Logger` is deliberately nil. Keeping that and the fuller `helper.go` one, and dropping the middle restatement, loses nothing.
2. **"a host with no OS keychain (Linux, containers)" appears five times** — `authsource.go` (twice), `expiry_aware_resolver.go`, `resolver_test.go`, `refreshlock_test.go`. It's the key insight, so worth stating once where the decision lives (`StoreManaged`) and letting the rest say "file-stored login".
3. **`ExpiryAwareResolver`'s comment is one ~230-character line** while the rest of the diff wraps at ~80, and it mostly narrates the routing the body shows.

Two that restate their own signature:

- `// Run reads one credential-helper request from in, resolves a credential via resolve, and writes the JSON response to out.` — those are the parameter names. The second sentence ("Returns an error on unparseable input or when no credential could be resolved") is the part worth keeping.
- `DescribeResolved` / `DescribeStoredCredentials` both open by restating their name before reaching the distinction that matters (OAuth login vs manual `auth set`).

## Not blockers — two questions

- **`staleCacheHint = 1 minute` on a permanently broken login.** Once a login needs `auth login`, every RPC past the minute re-spawns the helper and each spawn attempts its own `EnsureFresh` — a fresh network attempt per spawn per minute, N-way parallel. `warnCooldown` silences the message but not the attempts. Is a backoff worth it, or is the 1-minute pickup of a repaired login the more important property? I lean toward the latter, just want it to be a decision rather than a default.
- **`refreshLockTTL = 30s` break-open.** A refresh is two sequential network calls; on a bad connection 30s isn't unreachable, and breaking the lock then lets two processes spend the same rotated refresh token — the failure the lock exists to prevent. Deliberately tight to avoid wedging builds, with the double-spend accepted as the rarer risk?

## What I verified rather than assumed

- **`expiresLead` holds by construction.** It's `oauth.RefreshSkew / 2`, not a hand-picked duration, so the "MUST stay below `RefreshSkew`" rule can't drift — and there's a test pinning the consequence. Nice.
- **A zero `PATExpiry` can't produce a past `expires`.** I went looking because I saw a real machine today with `pat_expiry: 0001-01-01T00:00:00Z` (a flattened credential record). `Expiry: PATExpiry.Add(-expiresLead)` would be a *past* time there and `!cred.Expiry.IsZero()` wouldn't catch it. It's safe only because every success path out of `EnsureFresh` sets `PATExpiry` from `exchangeJWTForPAT`, which falls back to `now + defaultPATLifetime`. Worth knowing the guard depends on that fallback rather than on the `IsZero` check.
- **`filelock`'s "error means proceed unserialised" contract** is honoured at the one call site (`flow.go` debug-logs and continues). Unusual contract, correctly used.

## Context: the duplication here now has a ticket

`bazelcredhelper.newResolver` is step-for-step `configcommon.ExpiryAwareResolver.Get()` — resolve → serve as-is if not store-managed → refresh → serve stale on failure — differing only in return type and failure policy. **That's not a change request for this PR.** You couldn't have reused it: `RefreshFunc` exists to dodge a cycle (`internal/oauth` already imports `internal/config/common` via `internal/auth/store`) and it flattens `oauth.Credentials` to `(pat, workspaceID)`, discarding the `PATExpiry` this PR needs.

Filed as **[ACI-5274](https://bitrise.atlassian.net/browse/ACI-5274)** (under *ACI - Technical Improvements*), together with the wider inventory from [Ramil's #443 thread](https://github.com/bitrise-io/bitrise-build-cache-cli/pull/443#discussion_r3691058018): four precedence orders, four credential types, two mismatched provenance enums, and the five bugs that trace back to them. Your `StoreManaged()` change is a step toward that facade, so this PR moves in the right direction.

One process note: this branch is based on `29a66b1` (pre-#449), so its green CI ran without those changes. No file overlap and no textual conflict, but I'd like one green run on the merged result before we tag — which is what I'm waiting on anyway.

zsolt-marta-bitrise:
Note: decided to ship the other auth fixes I made until this is ready-to-be-shipped (see earlier AI comment)

internal/bazelcredhelper/resolver.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	var msg string
+	switch {
+	case errors.Is(err, oauth.ErrNotLoggedIn):
+		msg = "bitrise-build-cache: the stored token cannot be renewed automatically — run `bitrise-build-cache auth login`"
+	case errors.Is(err, oauth.ErrLoginRequired):
+		msg = "bitrise-build-cache: the Bitrise sign-in expired — run `bitrise-build-cache auth login`"
+	default:
+		// Transient (network, timeout): the stored token is probably still good.
```
zsolt-marta-bitrise, line 70: Found while testing this on a keychain-less Linux box: **a network blip during the refresh tells the user their sign-in expired and to re-run `auth login`.** The `default:` branch here can't be reached for a refresh failure.

`flow.go` relabels *every* `refreshJWT` error as `ErrLoginRequired` before `warnStale` ever sees it:

```go
refreshed, err := c.refreshJWT(ctx, creds.RefreshToken)
if err != nil {
    return Credentials{}, fmt.Errorf("%w (refresh failed: %w)", ErrLoginRequired, err)
}
```

so connection-refused, DNS failure and timeout all match `case errors.Is(err, oauth.ErrLoginRequired)`. Measured on the RDE — the two cases are textually identical:

```
# real WorkOS 400 invalid_refresh_token
stderr: bitrise-build-cache: the Bitrise sign-in expired — run `bitrise-build-cache auth login`

# IdP unreachable (closed port), stored token perfectly fine
stderr: bitrise-build-cache: the Bitrise sign-in expired — run `bitrise-build-cache auth login`
```

The serve-stale-and-continue behaviour is right in both cases — only the advice is wrong. Telling someone to re-authenticate mid-build when their token is fine and the network hiccuped sends them down the wrong path, and `warnCooldown` means it's the only thing they'll see for 10 minutes.

That wrap in `flow.go` is **pre-existing, not something this PR introduced** — I checked `e48b912`. But `warnStale` is new here and is the first code to act on the distinction, so it's the first place the conflation has a user-visible consequence. Either unwrap and skip the warning on a transport error, or have `flow.go` stop relabelling them. Your call whether that's in scope for this PR or a follow-up.

For completeness, everything else I threw at it behaved correctly:

| case | result |
|---|---|
| no credentials at all | exit 1, **stdout empty**, error names the doctor command |
| legacy static PAT (no refresh token) | served as-is, `ErrNotLoggedIn` warning, correct advice |
| expired file-stored login, IdP healthy | refreshed, rotated, persisted — `pat/jwt/refresh` all new |
| refresh refused (400) | stale token + 1-minute hint, warned once, **refresh token not burned** |
| 8 parallel helper spawns | exactly 1 refresh grant, all 8 got the same PAT |

The empty-stdout-on-failure case is the one I most wanted to see — a single stray line there breaks every build using the helper, and it's clean.

zsolt-marta-bitrise:
## Tested end-to-end on a keychain-less Linux RDE

Ran this branch (`8b3d612`) on a headless Ubuntu box with no secret-service, so the login lands in the config file — the exact host class the PR is about. `auth status` reports it as **"OAuth login (config file)"**.

First pass used a synthetic credential against a local stub IdP; second pass used a **real `auth login`** and the live WorkOS / Bitrise endpoints. Token fingerprints only below.

### Both refresh paths work against the real IdP

**JWT→PAT exchange** (backdated `pat_expiry`, JWT still inside the skew):

| | before | after |
|---|---|---|
| PAT | `e4239fde65` | **`00e468eb0f`** |
| JWT | `92a64370e9` | unchanged |
| refresh | `bb6141d6de` | unchanged |

Served `expires: 10:06:13` against a stored expiry of `10:06:43` — a 30s lead, exactly `RefreshSkew/2`. The refresh token is correctly left alone when the cheap exchange suffices.

**Full refresh-token grant** (backdated `jwt_expiry` too, forcing real WorkOS):

| | before | after |
|---|---|---|
| PAT | `00e468eb0f` | **`d1d4eeceb5`** |
| JWT | `92a64370e9` | **`e4f82400cc`** |
| refresh | `bb6141d6de` | **`093e52a787`** ← rotated |

WorkOS rotated the refresh token and it was persisted. Login healthy after every forced refresh.

### The regression is real

Same seed against `e48b912` (this branch's base, pre-fix):

```
# pre-fix — stale token, no cache hint, ZERO requests to the IdP
{"headers":{"authorization":["Bearer pat-OLD"]}}

# with the fix
{"headers":{"authorization":["Bearer pat-NEW"]},"expires":"2026-08-05T09:54:54Z"}
```

It never attempted a refresh. Worth stating because a passing test on the fix alone doesn't demonstrate there was a bug.

### Bazel end-to-end

`activate bazel` generated `--credential_helper=*.services.bitrise.io=bitrise-build-cache` (bare name, `$PATH`-resolved) with **no** literal `authorization` header anywhere in the bazelrc — `grep -c` returns 0. `bazel build` on Bazel 9.2.0 succeeded, no `UNAUTHENTICATED`, and results streamed to BES with `wait_for_upload_complete`. Since the template omits the literal header in helper mode, the only possible source of the bearer token BES accepted was the helper itself — so this exercises the whole chain, not just the protocol shape.

### Everything else I threw at it

| case | result |
|---|---|
| no credentials at all | exit 1, **stdout empty**, error names `doctor --fix --interactive` |
| legacy static PAT (no refresh token) | served as-is, `ErrNotLoggedIn` warning, correct advice |
| refresh refused (400) | stale token + 1-minute hint, warned once, refresh token **not** burned |
| 8 parallel helper spawns, all expired | exactly **1** refresh grant, all 8 served the same PAT |

The empty-stdout-on-failure case is the one I most wanted to see — a single stray line there breaks every build using the helper, and it's clean. The 8-way spawn is the lock doing its job with a real rotation on the line.

### One open item

The transient-vs-expired conflation in [this thread](https://github.com/bitrise-io/bitrise-build-cache-cli/pull/450#discussion_r3719306225): a network blip during refresh tells the user their sign-in expired and to re-run `auth login`. Measured — connection-refused and a real `invalid_refresh_token` produce byte-identical stderr. The wrap causing it is pre-existing in `flow.go`, and the serve-stale behaviour is right in both cases; only the advice is wrong. Not a blocker, your call whether to fix here or follow up.

Comments trimmed in `8b3d612` (−26/+12): the stdout rule went 4 places → 2, the keychain-less explanation 5 → 1 (kept on `StoreManaged`, where the decision lives). Lint and unit tests clean, PR CI green.

Also filed **[ACI-5274](https://bitrise.atlassian.net/browse/ACI-5274)** for the wider credential-resolution consolidation, so that stays out of this PR's scope.


[ACI-5274]: https://bitrise.atlassian.net/browse/ACI-5274?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

zsolt-marta-bitrise:
Moving detail out of the description into the thread, where it belongs under the [new PR-description shape](https://github.com/bitrise-io/claude-plugins/pull/94) — the body keeps what changes a review decision, the rest lives here. Nothing below is new; it's the original write-up preserved.

## The bug, as originally diagnosed

With `build --credential_helper=*.services.bitrise.io=bitrise-build-cache` in `.bazelrc`, builds work after activation and then break:

```
UNAUTHENTICATED: token validation failed: invalid Personal Access Token: unauthorized
ERROR: The Build Event Protocol upload failed: All 4 retry attempts failed.
```

`cmd/get/get.go` overrode the root `PersistentPreRun` with a no-op for spawn latency, which also skipped `hydrateStoredAuth` — the only caller of `oauth.EnsureFresh`. The helper served whatever was on disk, forever. Reproduced on v3.4.0.

Three related defects surfaced alongside it:

- **No `expires` in the response.** The EngFlow [`GetCredentialsResponse` schema](https://github.com/EngFlow/credential-helper-spec/blob/main/schemas/get-credentials-response.schema.json) defines an optional RFC 3339 `expires`. Without it Bazel falls back to `--credential_helper_cache_duration` (**30m** default on 8.4.2), so even a refreshing helper is only consulted every 30 minutes.
- **File-stored logins skipped.** `ExpiryAwareResolver.Get()` and `DescribeResolvedWith` both gated on `source == AuthSourceKeychain`. A host with no OS keychain stores its login in the config file via `store.SelectAuto` — so the resolver was a no-op precisely where it was needed, and `auth status` mislabelled the login as a plain PAT with no expiry.
- **Concurrent-refresh race.** Bazel spawns the helper N-way parallel and WorkOS rotates the refresh token per grant. Once the helper refreshes, several processes can each spend the same refresh token and the last writer persists one the others already burned.

## Four further defects found while reviewing the above

Each has a test that fails without the fix:

- **`EnsureFresh` did not re-read after a *failed* lock wait.** Timing out is precisely the case where another process is refreshing, so the credential loaded before the wait is the one whose refresh token has already been rotated away — and it was being spent anyway.
- **The multiplatform config's temp file was a fixed `.tmp` name**, so one writer could truncate the bytes another was about to rename into place.
- **`stop-proxy` unlinked the proxy pid file**, which is the lock file — the double-hold hazard described in the PR body.
- **The proxy's pid advertisement was consulted before the lock**, so a reader catching `WriteFile` mid-truncate concluded "not running" and started a second proxy.

## Other things worth knowing, none of which should change the review

- **A refresh failure serves the stored token rather than failing the RPC.** A stale token is a soft cache miss for Bazel; a non-zero helper exit fails the RPC outright. The stderr hint is rate-limited across the build's many spawns.
- **The resolver takes an `io.Writer`, not a `Logger`.** `log.NewLogger()` writes to stdout, which is the protocol channel — a logger anywhere on this path corrupts the JSON. `cmd/get/get_test.go` pins that stdout carries the response and nothing else.
- **`expiresLead` is derived as `RefreshSkew / 2`** rather than hand-picked, so the "must stay below `RefreshSkew`" rule can't drift; a test asserts it. A larger lead would make Bazel re-spawn the helper on every RPC while `EnsureFresh` still calls the PAT valid.
- **`stop-proxy` with nothing running now exits 0** ("No xcelerate-proxy is running") instead of failing with `read pidfile`. User-visible, so worth a second opinion.
- The `TokenSpentByADeadHolder` test asserts reuse is **1, not 0**: a process killed between the grant and the save leaves a token the IdP has already invalidated, and nothing local can recover it. The invariant is that it stays at one and the user is told to sign in again.

## Original numbers, v3.4.0 vs this branch

Same expired credential:

| | PAT served | `expires` |
|---|---|---|
| v3.4.0 | stale (`…2mOBtq`) | absent |
| this PR | refreshed (`…11p9pB`) | `2026-08-04T11:43:44Z`, persisted to disk |

Served PAT against `api.bitrise.io/v0.1/me` → **200** (was 401). `bazel build //...` on a real repo → 161 remote cache hits, BES upload clean. `auth status` now labels a file-stored login `OAuth login (config file) … token valid until …`; before it showed neither.

## Open item

A network blip during refresh tells the user their sign-in expired and to re-run `auth login` — `flow.go` relabels every `refreshJWT` error as `ErrLoginRequired`, so connection-refused and a genuine `invalid_refresh_token` produce byte-identical stderr (measured). The serve-stale-and-continue behaviour is right either way; only the advice is wrong. The wrap is pre-existing, not introduced here. Discussion in [this thread](https://github.com/bitrise-io/bitrise-build-cache-cli/pull/450#discussion_r3719306225).

## Follow-ups, not in this PR

- `feature-e2e-bazel-bitrisedc-credhelper` in `bitrise.yml` keeps `BITRISE_BUILD_CACHE_AUTH_TOKEN` set, so it only ever exercises the env branch and never the refresh path. Worth a variant that unsets it and seeds a file-store credential.
- Broader credential-resolution consolidation: [ACI-5274](https://bitrise.atlassian.net/browse/ACI-5274).
- A sweep for other places that now want a cheap crash-safe lock, in rough order of value: the **ccache storage helper has no singleton guard** and `os.Remove`s the incumbent's socket; **`/usr/local/bin` installs** do `LookPath` → `os.Create` → `io.Copy` on a binary that is immediately exec'd; the **xcodebuild/xcrun shims** are `O_TRUNC`-written while every build execs them; **Gradle `init.d` scripts** are written non-atomically while a daemon may be compiling them; and `makeSureCLIIsNotRunning` can be **deleted** rather than fixed, since `os.Rename` over a running executable is already safe.


[ACI-5274]: https://bitrise.atlassian.net/browse/ACI-5274?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

internal/bazelcredhelper/cooldown.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/gofrs/flock"
+)
+
+// claimCooldown reports whether every has passed since the last claim, re-stamping
```
Badlazzor: I don't quite get this half sentence. Don't we miss the what after "every"?

zsolt-marta-bitrise: oops :D seems like an artifact

zsolt-marta-bitrise: Nothing missing — `every` was the parameter name, so it parsed as Go-doc convention ("reports whether _every_ has passed") but reads exactly like a dropped word, which makes it bad writing regardless. Renamed to `interval`:

```go
// claimCooldown reports whether at least interval has passed since the last claim,
// re-stamping the marker when it has.
func claimCooldown(path string, interval time.Duration) bool {
```

Also fixed "when it returns true" → "when it has", which was describing the return value from inside its own doc.

Pushed in 07cff5b.

Badlazzor (approved):

PR #453, opened 2026-08-05
ACI-5273: doctor probes xcodebuild/xcrun resolve to the xcelerate wrapper

Badlazzor:
## Summary
- New doctor check `xcelerate-wrapper-path`: warns when `activate xcode` has installed the wrapper at `~/.bitrise-xcelerate/bin/{xcodebuild,xcrun}` but the current shell still resolves those to `/usr/bin` — the case where a user builds in the same shell they ran `activate` in and gets no cache while `Overall: ok` misleadingly says everything is fine.
- Gated on Xcelerate being activated; skipped otherwise. Same code path covers React Native since it activates Xcode.
- Uses `filepath.EvalSymlinks` on both actual and expected before comparing so installer / brew symlinks don't false-positive.

Fixes ACI-5273.

## Test plan
- [x] `make check` clean (pre-existing bazel `Test_enableForBazelCmdFn/No_envs_specified` failure on main aside).
- [x] Unit tests: activated + match OK, activated + mismatch warn, activated + not on PATH warn, not-activated skipped, symlinked wrapper OK.
- [x] Manual: run `activate xcode` in a fresh shell without sourcing rc, then `bitrise-build-cache doctor` — expect warn with `source ~/.zshrc` hint.

### DIFF
--- a/internal/doctor/check_xcelerate_wrapper_path.go
+++ b/internal/doctor/check_xcelerate_wrapper_path.go
@@ -1,0 +1,64 @@
+package doctor
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/bitrise-io/bitrise-build-cache-cli/v3/internal/paths"
+	"github.com/bitrise-io/bitrise-build-cache-cli/v3/internal/toolconfig"
+)
+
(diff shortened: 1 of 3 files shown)


zsolt-marta-bitrise (approved):
great idea, I also faced this issue before

PR #455, opened 2026-08-06
[ACI-5241] Report Build Cache health around each react-native run

zsolt-marta-bitrise:
[ACI-5241](https://bitrise.atlassian.net/browse/ACI-5241)

Follow-up to #444, which reported Build Cache health around each `xcodebuild`.

## TLDR

- One universal health gate (`internal/doctor.Gate`), instead of a per-wrapper copy.
- Both the `xcodebuild` wrapper and the new `react-native run` wrapper use it.
- RN checks a subset: the nested xcodebuild wrapper already covers the proxy and the cache errors.
- Gradle is not covered yet.
- Activation now creates the tool log dirs, so the first build after `activate` no longer warns about a missing one.

## The gate interface

Three hooks, so a wrapper only has to say *what* to check and *when*:

- `CheckAtStart(ctx, names)` — runs the wrapper's check set with the backend probe skipped (a network round-trip per build is too much overhead) and buffers the issues.
- `Recap()` — reprints them at the end of the build, a whole build's output later.
- `ProbeAuth(ctx)` — the probe an expired token would fail, run only when the analytics PUT fails; its verdict supersedes the buffered report.

Everything else stays with the wrapper: `xcodeDoctor` keeps its cache-on/off check sets and the proxy/CAS error reporting, `rnDoctor` keeps its check set. The report is always debug-logged, and the healthy path prints nothing.

RN's set excludes `xcelerate-proxy` (the nested wrapper starts it later, so it is down here by design) and includes `xcelerate-wrapper-path`, since RN reaches `xcodebuild` through the PATH shim — a shell that never picked it up means no Xcode caching at all.

Opt out with `--no-doctor` before the wrapped command, `BITRISE_BUILD_CACHE_SKIP_DOCTOR`, or `RunnerParams.SkipDoctor`.

## Verified on a clean macOS RDE

Real `activate react-native --xcode=true` plus a Swift package built through `react-native run -- xcodebuild`:

- Cold build, valid credentials — both gates all-`ok`, build green, RN parent + Xcode child invocations recorded ([RN](https://app.bitrise.io/build-cache/322a005426441b60/invocations/react-native/1f3b9b2e-9d77-4ceb-9ee2-e48f563e8fd0), [child](https://app.bitrise.io/build-cache/invocations/xcode/e702ece8-9cfc-4025-834c-f1b63cbd1693)).
- Warm rebuild — 100% reuse and zero warn-level gate lines ([RN](https://app.bitrise.io/build-cache/322a005426441b60/invocations/react-native/a088bee5-31d3-4545-be06-7629beac4eb1), [child](https://app.bitrise.io/build-cache/invocations/xcode/a2115bd6-a3af-4133-aeaf-4e91bafdbfbc)).
- Freshly activated — log dir present before the first build, both gates `ok log-dirs`.
- Bogus credentials — both gates report, the probe names it (`error auth-backend: … token rejected by Build Cache`), the child adds `The Bitrise cache reported 25 error(s)`, build still green.
- `auth clear` then rebuild — the RN half bypasses itself (`auth clear` empties the workspace ID `isReactNativeReady` needs) and the nested wrapper carries the report; an invalid *token* does not cause this, a missing *workspace* does.

[ACI-5241]: https://bitrise.atlassian.net/browse/ACI-5241?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bitrise_rn_config/bitrise.yml
```diff
@@ -939,6 +939,16 @@ step_bundles:
...
             cd _seek
             ../bitrise-build-cache-cli react-native run npx react-native build-android 2>&1 | tee -a "$RN_CLI_LOG"
+    - script:
+        title: Assert the wrapper ran its health check around the build
+        inputs:
+        - content: |-
+            set -exo pipefail
+
+            # `activate react-native --debug` above put the runner in debug mode,
+            # so the report lands in the wrapped command's log.
```
zsolt-marta-bitrise, line 949: Thinking whether we should add testing that auth errors are actually reported, we have that in the e2e xcode test, I didn't want to bloat it even more. LMK if you disagree

Badlazzor (approved):

PR #457, opened 2026-08-06
feat: ACI-5290 KV client conn pool + drop proxy ccSemaphore

Badlazzor:
[ACI-5290](https://bitrise.atlassian.net/browse/ACI-5290)

## TLDR

- KV client dials a pool of `max(2, NumCPU/6)` gRPC conns; round-robin picks one per RPC.
- Per-channel `chan struct{}` semaphore (size `NumCPU`) throttles inside the KV client; proxy `ccSemaphore` deleted (was double-throttle).
- `keepalive.ClientParameters{Time: 30s, Timeout: 10s, PermitWithoutStream: true}` on every conn.

## Why the 30s CAS waiter starved

Apple's `libToolchainCASPlugin.dylib` bakes in a 30s pool-waiter (`GRPCChannelPool.Configuration.ConnectionPool.maxWaitTime` at `0x1e630` = `30_000_000_000` ns) — not tunable via any exposed plugin option. Any queue > 30s between the plugin issuing an RPC and an HTTP/2 stream becoming available surfaces as `CAS operation failed: deadlineExceeded(connectionError: nil)`. Two chokepoints in the CLI created that queue: the xcelerate proxy's `ccSemaphore` capped concurrent RPC handlers at `max(2, NumCPU/6) * NumCPU` (20 on a 10-core Mac), and the KV client dialed a single `grpc.ClientConn` for all three sub-clients — every downstream RPC multiplexed over one HTTP/2 conn to BE. Fix mirrors the gradle plugin's `ClientBalancer`: one pool of connections, per-channel throttling inside the KV client, and the outer proxy semaphore removed so gRPC's own stream backpressure is the only limit.

## Testing

`xcodebuild build` on IceCubesApp, remote CAS fully populated (10 833 blobs / ~1 GB), identical workload each run:

| | Before | After |
|---|---|---|
| Wall time | 24m 10s | 1m 18.7s |
| Proxy blob hits | 10 757/10 757 (100%) | 10 833/10 833 (100%) |
| Xcode task hits | 1 820/1 896 (96%) | 1 896/1 896 (100%) |
| Cache-reported errors | 76 | 0 |
| Failed-upload lines in proxy log | 380+ | 0 |
| Per-session proxy log | 111 KB (retry noise) | 136 B |
| Post-run retries poisoning next session | yes | no |

`make lint` clean; `go test -tags unit -race ./internal/build_cache/kv/...` green (16 tests including 4 new for pool sizing, round-robin distribution, and stub-injection collapse).

[ACI-5290]: https://bitrise.atlassian.net/browse/ACI-5290?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/build_cache/kv/client.go
+++ b/internal/build_cache/kv/client.go
@@ -243,6 +243,8 @@ type writer struct {
 	response     *bytestream.WriteResponse
 	release      func()
 	releaseOnce  sync.Once
+	closeOnce    sync.Once
+	closeErr     error
 }
 
 func (w *writer) Response() *bytestream.WriteResponse {
@@ -279,13 +281,15 @@ func (w *writer) doRelease() {
 func (w *writer) Close() error {
 	defer w.doRelease()
 
-	var err error
-	w.response, err = w.stream.CloseAndRecv()
-	if err != nil {
-		return fmt.Errorf("close stream: %w", err)
-	}
+	w.closeOnce.Do(func() {
+		resp, err := w.stream.CloseAndRecv()
+		w.response = resp
+		if err != nil {
+			w.closeErr = fmt.Errorf("close stream: %w", err)
+		}
+	})
 
-	return nil
+	return w.closeErr
 }
 
 type reader struct {
--- a/internal/build_cache/kv/client.go
+++ b/internal/build_cache/kv/client.go
@@ -232,7 +232,7 @@ type writer struct {
 	offset       int64
 	fileSize     int64
 	response     *bytestream.WriteResponse
-	release      func() // nil-safe; releases the pool-entry semaphore once the stream drains
+	release      func()
 	releaseOnce  sync.Once
 }
 
@@ -286,7 +286,7 @@ type reader struct {
 	buf      bytes.Buffer
 
 	metadataReady chan struct{}
-	release       func() // nil-safe; releases the pool-entry semaphore on Close
+	release       func()
 	releaseOnce   sync.Once
 }
 
@@ -368,13 +368,17 @@ func (r *reader) Metadata() map[string]string {
 	return m
 }
 
-func (r *reader) Close() error {
-	r.buf.Reset()
+func (r *reader) doRelease() {
 	r.releaseOnce.Do(func() {
 		if r.release != nil {
 			r.release()
 		}
 	})
+}
+
+func (r *reader) Close() error {
+	r.buf.Reset()
+	r.doRelease()
 
 	return nil
 }
--- a/internal/build_cache/kv/client.go
+++ b/internal/build_cache/kv/client.go
@@ -36,10 +36,8 @@ const (
 )
 
 // Channel-pool sizing mirrors the Gradle plugin's ClientBalancer.
-var (
-	numChannels     = max(2, runtime.NumCPU()/6)
-	perChannelLimit = runtime.NumCPU()
-)
+func numChannels() int     { return max(2, runtime.NumCPU()/6) }
+func perChannelLimit() int { return runtime.NumCPU() }
 
 // AuthSource returns the credentials to use for a single RPC. Implementations
 // may cache and refresh transparently; kv.Client re-reads on every call.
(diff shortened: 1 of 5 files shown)


internal/xcelerate/proxy/proxy.go
```diff
@@ -308,9 +299,6 @@ func (p *Proxy) GetSessionStats(_ context.Context, _ *emptypb.Empty) (*session.G
 }
 
 func (p *Proxy) Get(ctx context.Context, request *llvmcas.CASGetRequest) (*llvmcas.CASGetResponse, error) {
-	p.ccSemaphore <- struct{}{}
```
zsolt-marta-bitrise, line 311: Shouldn't we close the streams?

internal/build_cache/kv/pool_internal_test.go
```diff
@@ -0,0 +1,137 @@
...
+		got := counts[i].Load()
+		assert.GreaterOrEqualf(t, got, int64(200), "entry %d saw %d calls, want >= 200", i, got)
+		assert.LessOrEqualf(t, got, int64(300), "entry %d saw %d calls, want <= 300", i, got)
+	}
+}
+
+// Injecting BitriseKVClient (or CapabilitiesClient) collapses the channels to one
+// entry with no semaphore, so pickChannel never blocks — even under load that
+// would otherwise exhaust perChannelLimit.
+func TestBuildPool_InjectedStubIsUnthrottled(t *testing.T) {
```
zsolt-marta-bitrise, line 95: Do we use this? I thought it's for testing setup but seems lik this is meant for some production code? Do we inject the one-off clients anywhere?

internal/build_cache/kv/client.go
```diff
@@ -127,22 +164,77 @@ func NewClient(p NewClientParams) (*Client, error) {
...
+			for _, ch := range channels {
+				_ = ch.conn.Close()
+			}
+
+			return nil, fmt.Errorf("dial %s: %w", p.Host, err)
+		}
+
+		channels = append(channels, &channel{
+			conn:               conn,
+			sem:                make(chan struct{}, perChannelLimit),
```
zsolt-marta-bitrise: I like this interface of putting the clients in the channels and the reader/writes one layer down. Simple to use with our current proxy setup

zsolt-marta-bitrise (commented):
Thank you for using the new TLDR PR description format! Please check the CI too

internal/xcelerate/proxy/proxy.go
Badlazzor: Good catch — `upload.go` only called `kvWriter.Close()` on the happy path, so error branches (`io.Copy` fail, `AlreadyExists`, `Unauthenticated`) skipped both `stream.CloseAndRecv()` and the sem release. Fixed in 1dd540f: `defer kvWriter.Close()` right after `initiatePut` succeeds; `writer.Close()` is now `sync.Once`-guarded on both the stream close and the release so the deferred + explicit calls dedup on the happy path. Download side already had `defer kvReader.Close()`.

internal/build_cache/kv/pool_internal_test.go
Badlazzor: The injection path is test-only — `upload_test.go` / `download_test.go` / etc. wire just a mock `BitriseKVClient` so `buildChannels` skips dialing. This test locks that collapse contract (`len==1`, `sem==nil`, `conn==nil`, concurrent acquire/release never blocks) so if someone changes the shape of the test shortcut, it fails loudly here rather than downstream in a mock-based test.

internal/xcelerate/proxy/proxy.go
zsolt-marta-bitrise: What about download?

Badlazzor: As stated in the above comment it was already good.
download.go:115 already has defer kvReader.Close() right after initiateGet

zsolt-marta-bitrise (approved):

PR #459, opened 2026-08-07
[ACI-5274] Consolidate credential resolution behind one auth facade

zsolt-marta-bitrise:
[ACI-5274](https://bitrise.atlassian.net/browse/ACI-5274)

> **Review commit by commit** — 9 commits, each green on its own. Depth per commit in the guide below; three of them are worth real attention and the rest are skims.

## TLDR

- One resolver, `internal/auth/live`, replaces seven precedence orders and four copies of resolve-then-refresh.
- Four credential types collapse to two; `AuthSource` (6 values) and `store.Kind` (2) collapse to `Origin{Backend, Provenance}`.
- `internal/oauth` moves under `internal/auth`; `config/common` leaves the auth graph entirely.
- Layering is enforced, not documented: two import invariants via `depguard`, two symbol invariants via `scripts/lint_arch.sh`, both under `make lint`.
- `pkg/common` and `pkg/reactnative` resolve through the facade too, so no consumer reads a credential off a config struct.
- Fixes five live bugs: unrefreshed file-stored logins, every token refresh erasing the display name, `activate` dropping the credentials block, `auth token` serving an expired PAT, and the invocation PUT disagreeing with the doctor.
- New reference: [`docs/auth.md`](docs/auth.md).

## Review guide

| # | Commit | Depth | What to look at |
|---|---|---|---|
| 1 | `39052c0` docs: the target layering | **Read first** | The design the other eight implement. `docs/auth.md` is the map — the diff is much easier with it in hand. |
| 2 | `c9d51f9` `internal/auth` leaf types | **Read closely** | `Credential` vs `TokenSet`, the one-way conversion, `Origin`. Everything downstream follows from these ~200 lines. |
| 3 | `7969f29` move `oauth` under `auth` | Skim | Pure path rename, 11 files. |
| 4 | `884eea1` `store` takes `isCI bool` | Skim | 4 signatures, 4 call sites. |
| 5 | `4ec76bd` collapse the storage types | Normal | Where `keychain.Credentials` and `oauth.Credentials` merge. The `Username`-drop fix is here. |
| 6 | `879aa03` the `live` resolver | **Read closely** | The facade itself, plus the 16-combination precedence table that is the regression net for the whole PR. |
| 7 | `0b790bd` migrate every consumer | Skim **by package** | ~90 files, almost entirely mechanical. Reading it line by line is a poor use of time; check one file per package and move on. |
| 8 | `0e21127` comment trim + docs | Skim | No code change. |
| 9 | `8464b05` the review passes | **Read closely** | Every regression the reviews caught in commits 2–8, each with a test confirmed to fail against the unfixed code. Also finishes the job: `pkg/common` and `pkg/reactnative` now resolve through `live` instead of reading a credential off a config struct. Collapsed into one commit so the history stays readable. |

If you only have time for three: **2, 6, 9**.

## The two credential types

The split worth pushing on is `auth.Credential` vs `auth.TokenSet`, because it is what makes the rest safe.

`TokenSet` is the persisted record — token, workspace, display name, and the OAuth refresh machinery. `Credential` is what a caller needs to make one authenticated call: token, workspace, username, expiry. Twenty-plus consumers take `Credential`; none of them needs a refresh token, and handing one over is how it ends up in a log line or a rendered init script.

The conversion is `TokenSet.Credential()`, and **there is deliberately no reverse.** Writes load the stored record, mutate the two fields they own, and save it back. Constructing a record from a boundary value is exactly what the old `mergeActivateCreds` did when it destroyed a refresh token on a real machine (#449); here it does not compile. The same reasoning killed the third and fourth types: `keychain.Credentials` and `oauth.Credentials` were the same seven fields under two names, bridged by a conversion that silently dropped `Username`.

`Origin` exists because backend and provenance are independent axes. The config file holds both an `auth login` credential and the analytics `authConfig` block — same backend, different provenance — and only the pair distinguishes them. `Origin.StoreManaged()` is what gates refresh; the old code tested `source == AuthSourceKeychain`, which is why file-stored logins were never refreshed.

## Testing

- `make lint` (golangci-lint + `lint_arch.sh`) and `go test -race ./...` clean; Bitrise pipeline green; all 9 commits build individually.
- Ran on a real macOS box against a live workspace: `auth login`, `auth login --storage=file`, `activate xcode/c++/bazel`, `doctor`, the Bazel credential helper.
- #450 proved directly: forced `pat_expiry` into the past, and the CLI refreshed the **file-stored** login — token `…Q_3Cenfr` → `…g_1g8tRR`, new expiry, display name intact. That path never refreshed before.
- Negative controls: the `depguard` rules were confirmed to fire by planting a violating import in `internal/auth` and in `auth/store`; each regression test in commit 9 was confirmed to fail against the unfixed code.
- Not covered on hardware: the keychain→file fallback merge needs a broken keyring, so it is unit-tested only.

One release note: when env vars **and** a stored login both exist, the invocation PUT now sends the env credential. That is the #444 fix — `ReadConfig` resolves instead of reading its own config file — and the one case where a machine's effective credential changes.

[ACI-5274]: https://bitrise.atlassian.net/browse/ACI-5274?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Badlazzor (approved):
The only thing I found is a small testing gap:

happy-path SaveWithFallback invariant (keychain writes, file untouched) only implied. Add case: prime file, ResolvePinned with env credential + healthy keychain, assert file untouched. Guards against future flip to SaveExclusive* silently wiping user data.

PR #460, opened 2026-08-07
fix: ACI-5291 imply --cache when --cache-push is set on activate gradle

Badlazzor:
[ACI-5291](https://bitrise.atlassian.net/browse/ACI-5291)

## TLDR

- `activate gradle --cache-push` (without `--cache`) silently disabled the remote cache — init script wrote `Usage: none` and `gradle.properties` wrote `org.gradle.caching=false`.
- `NormalizeParams` now flips `Cache.Enabled=true` whenever `Cache.PushEnabled=true`, called at the top of `Activate` so both the init-script inventory and `UpdateGradleProps` see the mutation.

## Why the normalizer sits in `Activate`

`Cache.Enabled` and `Cache.PushEnabled` are consumed by two independent writers — `templateInventoryProvider` (init script) and `UpdateGradleProps` (gradle.properties). If normalization ran only in one of them, the other would drift. Placing it in `Activate` before either writer runs keeps both outputs consistent from a single source of truth, and the params struct stays a plain data holder without hidden getter logic.

Related follow-up filed: [ACI-5292](https://bitrise.atlassian.net/browse/ACI-5292) — same shape but on benchmark baseline; out of scope here.

## Testing

- `make lint` — 0 issues
- `make test-unit` — green (pre-existing `Test_enableForBazelCmdFn/No_envs_specified` flake on main, unrelated)
- New end-to-end `Test_Activate_CachePushImpliesCacheEnabled` — verified to FAIL when the `NormalizeParams(&params)` call is removed

[ACI-5291]: https://bitrise.atlassian.net/browse/ACI-5291?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ACI-5292]: https://bitrise.atlassian.net/browse/ACI-5292?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/config/gradle/activate_for_gradle_params.go
```diff
@@ -70,12 +70,23 @@ func DefaultActivateGradleParams() ActivateGradleParams {
 	}
 }
 
+// NormalizeParams enforces param invariants before any consumer (init-script
```
zsolt-marta-bitrise: This is an unnecessary comment, it's an obvious method. Please use the PR skill to create the PR as it will trim the comments + use the TLDR format

Badlazzor: Sorry it was meaningful for me, I suppose this is a bit subjective.

zsolt-marta-bitrise (approved):

zsolt-marta-bitrise:
[//]: # ([skip ci])



## Summary
Fix cache activation when --cache-push is set without --cache. Added NormalizeParams() to ensure Cache.Enabled is true whenever Cache.PushEnabled is true, called early in both Activate() and TemplateInventory() so both init-script and gradle.properties outputs remain consistent.


## Walkthrough
| File | Summary |
|------|---------|
| internal/config/gradle/activate.go, activate_for_gradle_params.go | Added NormalizeParams() function and calls at entry points to ensure cache is enabled when cache-push is enabled |
| internal/config/gradle/activate_test.go, activate_for_gradle_params_test.go | Added and updated tests including new end-to-end test for cache-push implying cache-enabled |

